PR #6, opened 2019-01-03
ECIP-1047: Reduce Gas Limit to 1 Million

pyskell:
## Title

    ECIP: 1047
    Title: Reduce Gas Limit to 1 Million
    Author: Anthony Lusardi (ethereumclassicanthony@gmail.com)
    Status: Draft
    Type: Informational
    Created: 2019/1/3
    
## Abstract

All Ethereum-based chains face a major issue where blockchain bloat has the potential to drastically increase sync times and make it difficult to impossible for users to fully validate their copy of the ETC blockchain.

To mitigate this issue, this ECIP is a request that miners target a gas limit of 1 million which is an 8x reduction from the current limit of ~7.95 million; and well above the average gas size of blocks on the ETC blockchain.

## Motivation

Over the past year the ETH blockchain has grown substantially in verification time from less than 1 day in June 2017 to [4+ days in October 2018](https://twitter.com/lopp/status/1056533134760656898) and growing; even with considerably powered hardware and recent client optimizations. With maximum gas usage it's reasonable to expect that sync times over the next few years will grow to the point where a normal users and even some businesses cannot fully validate the chain, leading to node centralization and the need for trusted intermediaries. 

We have the benefit of hindsight and lower utilization but will face exactly the same problems if we do not take steps to address blockchain bloat.

## Specification

Miners can start reducing their gas limit to 1,000,000 by adding the below parameters to their clients. This change will happen over time with each block reducing by a maximum of 1/1024th of the gas of the previous block. More details on `gasLimitBoundDivisor` [are here](https://wiki.parity.io/Chain-specification).

### Geth
`--targetgaslimit 1000000`

### Parity
`--gas-floor-target 1000000 --gas-cap 0`

### Mantis
`TBD`

### Code Changes
A more robust method to decreasing gas target is to specify a default within each ETC client. Potential changes are detailed below:

- Parity-Ethereum, Mantis, and Multi-Geth will require a small refactor such that the default is specified in a config file, rather than hard coded as is presently the case.
- Geth-ETC, being a single-network client, can simply update the value [specified here](https://github.com/ethereumclassic/go-ethereum/blob/1a8b71b8f3e03775303182df7f446a93234ef89c/core/block_validator.go#L37)


oberondelafay:
Sorry but this is horribly misguided and just plain wrong.
Centralization is the nature of all blockchains where there is any subsidy offered by the network.
  
Centralization occurs because of economies of scale.  
Miners will consolidate and centralize based on their ability to mine efficiently.  This is why Bitcoin was so heavily Chinese years before blockspace became an issue.  

You're cutting your miner's reward per block.  This will cause miners who cannot mine profitably to drop out, further increasing centralization at a time where prices are already depressed and thus it is most dangerous.  

But I bet you're thinking...
"It could drive gas prices up and this will be enough to subsidize the miners".  That might happen for a short time.

But what you're actually asking here is to lower the utility of ETC for companies that would like to pay for and use that excess capacity for business purposes.  In otherwords you're asking to reduce blockchain capacity and this will trigger a cryptokitties style problem at a time that ETC doesn't even have a killer app yet.  This means that companies that were considering ETC because it has the ability to scale to meet their demands will now be forced to reconsider their investment simply from the threat that you would reduce the maximum block size to support the right to a timely full blockchain sync for someone in Nigeria with 2 tin cans, some string and a floppy disk.  

I can't speak for everyone else, but I know this proposal has just caused my company to put a full stop to all consideration of bringing what we had in development to ETC, which is doubly sad because we were looking at ETC precisely because ETH cannot scale to meet our needs due to exactly this.

What's your reasoning?
Because a bigger blockchain with bigger blocks makes it harder for people to do a full sync?
How about instead of limiting the utility and future uses by restricting what the users can put into the paid for capacity, we instead work to build a more efficient system for syncing the chain for people that want to run a full verifying node?  This can be done without shrinking block sizes, just by more efficiently utilizing the resources we have and making the chain more widely available from better connected systems.

oberondelafay:
Adding a small subsidy for running a fully verifying node would encourage people to run more fullnodes and completely eliminate all the problems you've mentioned.  Can we get an ECIP for that?

pyskell:
The alternative is that ETC is not independently verifiable by most users after a few years of full usage.

This is also entirely reversible by miners if other improvements are made.

realcodywburns:
you can copy and paste the template to submit your ecip

oberondelafay:
Thanks, I may actually do that.  Do you think it would make sense to add it to the treasury ECIP?
Meantime, I vote that lowering the gas limit at all is dangerous and short sighted and lowering it to 1M is just slow suicide.  Personally I think it should be raised to whatever amount the network can handle in the block time.  Then use the excess fees to subsidize chain hosting nodes.  But maxing it out would be a good start, it would attract a lot more business users.

saturn-network:
In case this ECIP passes and gets introduced to ETC we might indeed see gas price skyrocket in value. You may want to use this time of big empty blocks to mine some [GasTokens](https://forum.saturn.network/t/gastoken-tokenize-gas/2361).

oberondelafay:
Again this would be a less than desirable outcome for everyone in the long run.  Instability in the price of gas makes the platform far less attractive for anyone interested in deploying serious use cases.  It's one thing to mop up supplies and to start driving the ETC price up.  All you need are a few good killer dapps.  But once you start messing with the per transaction cost in terms of the unit of account (units of ETC required to get the transaction in), then it really loses its luster.

oberondelafay:
Cross posting my comments from Reddit...

Just following up. It's been 21 hours since I last updated.

I fell asleep sometime during syncing and woke up to a fully synced node, but it wasn't really syncing well when I went to sleep.

So most defiantly less than a day but more than 12 hours to go from 0 to a fully synced and full verification node on modern equipment.

I do have a 1 Gb fiber connection on a static IP, and my computer has 8 GB of RAM a Quad Core processor and a 1TB SSD.
It was high end when I bought it, but it's probably mid-range now.
But this is a 2 year old laptop.

The internet is nothing special either. Where I live 1Gb fiber is the default and considered mid range, so I pay about $50 per month for internet access, yet already I can go to 10Gb for $200 per month and they are discussing lowering prices already.

This is in podunk Utah, https://www.utopiafiber.com/residential-pricing/ I can only imagine more developed areas have better options.

I forgot to mention that I'm connected to my router via WiFi so my connection throughput speed is limited to 100MB/s to allow the wife and kids to stream their Netflix and Youtubes and Facebooks comfortably.

So I ask you. Is there really a problem you're trying to solve here? People that want to run a fully verifying node are going to either have the resources to do so, or have a business reason to acquire the resources to do it. They can do it now with the larger more spacious blocks having been the default for years and do so on what amounts to mid-range commodity hardware and there is no reason at all to expect that innovation and minimum expectations of service will not continue into the forseeable future.

In the meantime, shrinking blocks at this time literally rang the early warning alarm at my work and I now have to explain why moving to ETC may still be a viable idea. Understand, we have a functional product in testing on Rinkeby and will have to put in a lot of upfront effort to port to ETC because the EVM is very stale.

The most important thing to remember is that this gas limit per block represents a maximum utility. A blockchain to a business is worth not one dime more than the maximum amount of work they can put it to per unit of time. If you reduce the maximum size, then you reduce the utility value which then makes other chains much more attractive.

I'm the one talking because I care. But I can promise you that for every person who vocalizes a concern there are hundreds who will look, go "meh" and move to something else. It mostly depends on how far along they are in the development cycle. A true enterprise app has a lifecycle with a minimum of 2 year and an average of 5 years. These are enterprises who are early in their cycle and will simply overlook ETC if this comes to pass. That means less users in the future and of course that will make it much easier for those users to fully verify.

So it's your choice.

whilei:
@pyskell 

> All Ethereum-based chains face a major issue where blockchain bloat has the potential to drastically increase sync times and make it difficult to impossible for users to fully validate their copy of the ETC blockchain.

All decentralized and redundant databases are like this. This not a "bloating problem." This is _scaling behavior_. Indeed, assuming constant utilization, storage, consensus, and verification mechanisms, the longer the chain grows, the more space and time it will occupy. More space and time demands, well, simply that - more space and time; but "impossible" is quite a stretch from "4+ days." Google has been "syncing" my mother's photo library for 8 full days now.

> Over the past year the ETH blockchain has grown substantially in verification time from less than 1 day in June 2017 to 4+ days in October 2018 and growing; even with considerably powered hardware and recent client optimizations. 

Let's be clear about distinguishing "verification" and "sync." They're not the same thing. Yes, sync _includes and depends on_ verification, but it also includes peering, download, and storage systems. A sync process can use different verification mechanisms, ala fast, warp, light, archive, etc., and it's normal for a "sync" to actually use a combination of more than one strategy (or rather, for a stategy to include more than one verification mechanism). While verification does take time, so does downloading 100GB from a bunch of unpredictable peers that you don't trust.

> With maximum gas usage it's reasonable to expect that sync times over the next few years will grow to the point where a normal users and even some businesses cannot fully validate the chain, leading to node centralization and the need for trusted intermediaries.

Here's the disconnect for me. How does target gas usage per block have _anything to do_ with sync or storage space? _If we assume constant transactions hitting the chain and a gas limit that is not prohibitive to large contracts_, then they will occupy the _same amount of space and time_, no matter whether they're included in 4 blocks or 14. So I have to assume that you're advocating _fewer and smaller transactions in general_.

Which to me sounds like: Contracts and transactions are overrated anyways (hooray! we're "lower utilized"!), let's _discourage them_. By reducing normal capacity for utilization, we'll ward off a growing blockchain. Miners will be happy because sync (a one-time startup cost) will be fast and small, and won't have to process as many of those pesky transactions and contracts that would earn them gas (a recurring source of revenue).

Now, as economically crazy as I've made this sound, I can actually see your point. If the main intention for ETC value is $ETC/$USD rates in the short term, then this makes sense. "Faster sync! Get your faster, smaller sync here! Fewer and smaller transactions for miners to process so you can get right to the profitable block-mining! Earn your block rewards today! No one's developing any good Dapps? They don't want to use a chain that discourages them from building gas-heavy, complex contracts? Don't worry about gas revenues, we've got block rewards and a lightweight, under-utilized, chain!" Ohh... but wait. Long term. Block rewards are disinflating, scheduled to programmatically flatline, eventually shifting the revenue stream for miners entirely to the gas earned from transaction processing.

Maybe the core value for ETC and Ethereum-esque chains _won't_ come from processing _a lot of_ or _heavy_ transactions. In fact, at this point, I personally anticipate in the long term relatively few transactions hitting the chain, precisely because they will be prohibitively expensive except for the real valuable use-cases that demand it. But I don't know. 

But what I really don't know is __why you're ~~talking about~~ _advocating_ a single miner strategy here?__ Why are you trying to nudge market prices that miners can and will decide for themselves? Which begs the question, what do you have to gain? 

Do you really care this much about conveniently syncing the whole chain on your laptop? If so, I can encourage you to look into external drives and maybe recommend planning a few days before you're fully synced, or suggest that you contribute to one of the many discussions and undertakings surrounding _actual solutions to scaling_ instead of "let's discourage use." 

Or is there some other reason? Do you _not_ have Dapps that you'd like to see the chain handle expediently? Are you a miner who'd like to pump ETC in the short term with shortsighted "benefits" at the cost of undermining an immature-but-growing ecosystem that will support the network in the long term?

BelfordZ:
image

AFAIK miners should be the ones adjusting this based on what will make them most profitable, while users are free to incentivize miners to move up the block gasLimit. This value is really for miners to come up with on their own, and to optimize their mining software such that it meets the *current* demand of the network.

whilei:
Sure, thanks for saying something. Maybe walking a little close to the :fire:  

@pyskell I hope this hasn't offended, not meant to. It just seems very bizarre to me.

drd34d:
High gas limit can be perceived as an issue but there's good reasonings for it. Basicly It's up to the mining pool operators (>90% of the miners don't even care). There might be a better chance to lower it by reaching out directly to the operators.

But on the topic of chain "bloat"... 
Does anyone have a good reasoning for not **desincentivizing** empty blocks?

saturn-network:
disincentivizing empty blocks == incentivizing spam transactions

realcodywburns:
can you open an issue or thread on the forum for discussion and add a `discussions-to:` link in the header so it can be merged and tracked per the readme

ECIPs/ECIP-1047.md
```diff
@@ -0,0 +1,39 @@
+## Title
+
+    ECIP: 1047
```
realcodywburns: needs a `discussions-to:` header field

pyskell:
@whilei The analogy only covers about 80% of what this is since the shipping company does not actually own the warehouse where they store their trucks. The warehouse is a shared resource and every participant needs to make their own. I suggest avoiding analogies as they likely won't properly encapsulate the nuances of this situation.

The suggestion here to reduce the gas limit is because not doing so ultimately leads to centralization and difficulty syncing for participants of the network. 8 million gas in particular is quite substantial and it takes only 1 year of full blocks to create a weeks worth of syncing on typical hardware. We're seeing this on ETH along with storage sizes of ~130GB.

The core of my argument here is that blockchains in general socialize the cost of syncing nodes onto future participants and that the current gas limits, as evidenced on ETH, are not scalable long term.

P.S. No offense taken, I didn't even read the pre-edit post so I'll stay blissfully unaware of what may have offended me 😄 

@realcodywburns Done. Assuming we'll move all further discussion there?

zmitton:
Hi, wanted to voice my support for this ECIP (actually I don't even think it needs to be an ECIP because it's not a software update, just a miner config choice). 

I've been running Ethereum full nodes since mainnet was launched, and it's definitely out of hand today. I recently paid over $1000 to put a 2TB SSD in my macbook. As I write this (as per usual) my machine is running Bitcoin and Ethereum clients. Here is what it took to make that happen on an i7 with 16GB ram:

Bitcoin: ~3 days (on/off usage)
ETH: ~2 months

It's actually worse than that. I'm only running a "fast sync" on the ETH node (default), and it turns out, that I don't have access to older transactions. They are simply not on my machine. When I query the RPC for them it returns `null`. Also I became skeptic that my node was even working correctly (because the "latest" block would return blocknum: 0 for the full 2 months).

I'm an Ethereum developer, most Ethereum dev teams do not even run full nodes (archive). This is a huge problem, especially for Ethereum Classic whose existence is *only possible* because of full node security. I say this because, If you run only a fast sync, you are not validating all transactions, and would not catch the invalid DAO refund transaction. 

Ethereum classic has a huge reason for existing, and that reason must appreciate the importance of full nodes. 

ETC's main changes so far (beond the DAO debate) have been very attractive to me: (1) capping the supply. (2a)Commiting to PoW, and (2b) removing the difficulty bomb.

Having a reasonable sized full node would be yet another huge improvement over (its friend and competitor) Ethereum, and enable more users to develop on it.

I hope I've convinced some people, but really we only need to convince miners. Let's talk about the economics. 

Over the past year and a half, bitcoin proved to us that smaller will blocks result in larger fees for miners (*not* the other way around as was previously argued) as payments of its kind are very inelastic. In late 2017 Bitcoin hit its ceiling and tx fees skyrocketed to 30$. This "transaction traffic" had exponential properties similar to traffic patterns of cars. As Pierre Rochard had speculated years before, transactions could handle a much higher fee than they were being priced on bitcoin. In this sense the miners can nearly optimize fees by finding the inflection point where substitution starts happening and then raise the blocksize slowly to follow that threshold. ([see this discussion](http://bitstein.org/blog/pierre-rochard-and-gavin-andresen-discuss-the-block-size-limit/)) between Gavin and Pierre. Ethereum classic would be the first blockchain to experiment with these economics and could pave way in this regard (because bitcoin has a fixed limit, and Ethereum just follows what the foundation recommends, which is likely much higher than the market would find). 

There are many more benefits I won't touch on, but the tldr is that the benefits far outweigh the drawbacks.

pyskell:
100% agree with @zmitton, the current high gas limit is unscalable based on hard evidence of it in the wild. 

Lowering it is the prudent approach to prevent a catastrophic situation where syncing is unfeasible for most or nearly all participants. As efficiency in processing and hardware improves it can then be raised.

As mentioned by @zmitton this ECIP represents a recommendation to miners and changes no consensus guarantees about the network (it's a simple client configuration flag).

I'd also like to point out that at the time of the hard fork the gas limit was 4.71 million where it stayed until ETH recommended a gas target of 8 million for their network. Then in September it was made the default in parity for all blockchains (paritytech/parity-ethereum/pull/9564). This change seems to have caused the increase as we were still at ~4.7 in August (http://gastracker.io/block/6426100) and it started rising in September, jumping up 25% already by end of October (http://gastracker.io/block/6726100).

DonaldMcIntyre:
I support this ECIP because we have the advantage of hindsight and blockchains as Ethereum and Bitcoin have huge node churn rates with little new node growth, which has resulted in large node count losses in both networks.

Long term bandwidth growth in the US for example is about 17%, but blockchains are growing at a much larger pace which only means that the synch times for regular users (this is the key, that these chains need regular users to easily run nodes as well) is only getting longer which disincentives peole of running nodes in regular machines. 

On the other hand, disk space for regular users is also a constraint. A 2017 Macbook Pro, a very advanced kind of machine on average, for example, has a disk size of 250 GB, but Ethereum archive node is 2.2 TB, and normal node is 180 GB. A bitcoin full node (archival) is 220 GB.

It is a fallacy for those who think that "people can just connect a pen-drive and solve the disk size problem" because that kind of action is typical of a more advanced user, not regular users who just want convenience.

As a separate issue related to this, UX is terrible in ETH and ETC wallets and node software. It is incredibly easy to download Bitcoin Core and Bitcoin Knots and just synch the chain for a regular user. For it's nightmare to download Classic Geth or Multi-Geth and try to set them up as a non-technical user.

What more steps do we have to do to complete this ECIP and convince all miners and node operators to adopt it?

[if this is a done deal and already approved, and no need for more efforts, then pls disregard this comment]

zmitton:
@DonaldMcIntyre seems it's mostly about convincing miners or mining pools to use the geth config flag specified by @pyskell (`--targetgaslimit 1000000`)

DonaldMcIntyre:
I’m in to help in this campaign contacting miners

zmitton:
continue discussion https://github.com/ethereumclassic/ECIPs/issues/14

PR #72, opened 2019-04-21
ECIP-1060: Specify the Cliquey proof-of-authority engine.

soc1c:
I propose to upgrade the Kotti testnet to Cliquey (a.k.a. "Clique v2") to make it more robust and prevent it from getting stuck in future.

References https://github.com/ethereum/EIPs/pull/1955 https://github.com/goerli/eips-poa/issues/12

### DIFF
--- a/ECIPs/ECIP-1060.md
+++ b/ECIPs/ECIP-1060.md
@@ -1,0 +1,413 @@
+---
+ecip: 1060
+title: Cliquey proof-of-authority consensus protocol
+author: Aidan Hyman (@ChainSafeSystems), Talha Cross (@soc1c)
+discussions-to: https://github.com/goerli/eips-poa/issues/12
+status: Draft
+type: Standards Track
+category: Core
+created: 2019-04-19
+requires: 1048
+---
+
+## Simple Summary
+
+This document proposes a new proof-of-authority consensus engine that could be used by Ethereum testing and development networks in the future.
+
+## Abstract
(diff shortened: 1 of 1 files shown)


ECIPs/ECIP-1060.md
```diff
@@ -0,0 +1,413 @@
...
+}
+```
+
+By using this list, for convenience, the `extraData` field of the genesis block only has to contain the 32 bytes of **`EXTRA_VANITY`**. The client automatically converts and appends the list of signers to the block `0` extra-data as specified in EIP-225 at checkpoint blocks (appending `SIGNER_COUNT*20` bytes).
+
+### Sealing
+
+For a detailed specification of the block authorization logic, please refer to EIP-225 by honoring the constants defined above. However, the following changes should be highlighted:
+
+* Each singer is **allowed to sign any number of consecutive blocks**. The order is not fixed, but in-turn signing weighs more (**`DIFF_INTURN`**) than out-of-turn one (**`DIFF_NOTURN`**). In case an out-of-turn block is received, an **in-turn signer should continue to publish their block** to ensure the chain always prefers in-turn blocks in any case. This strategy prevents in-turn validators from being hindered from publishing their block and potential network halting.
```
YazzyYaz: Yes! We ran into issues of offline validator nodes hindering the whole network, so allowing in-turn signer to continue seems like a blessing.

PR #74, opened 2019-05-04
Update ECIP-1060 to align with EIP-1955

soc1c:


soc1c:
Needs some work. :hourglass_flowing_sand:

PR #81, opened 2019-06-06
ECIP-1061: Atzlán Hardfork Meta

soc1c:
### Atzlán Hardfork Meta

Enable the outstanding Ethereum Foundation _Istanbul_ network protocol upgrades on the Ethereum
Classic network in a hard-fork code-named _Atzlán_ to enable maximum compatibility across these networks.

Aztlan-Legendary-Aztec-Land


zmitton:
These names sound all a bit sound too similar. how about _Indus_, an ancient civilization and one of the longest rivers through china, india and pakistan.

I think we should oppose the following EIPs from Istanbul:
  - EIP-1057: ProgPoW, a Programmatic Proof-of-Work
  - EIP-2029 and EIP-2031 State Rent

Ethereum has not finalized which of the EIPs will make it in to Istanbul, but other than the above, I think we should implement whichever ones they end up implementing.

TokenHash:
If Agharta will be in ~Dec 2019, I propose Atzlán for September 2020.

Rationale: To minimize risks and expanding the "argument surface".

Separately, I agree with @zmitton re: 

`I think we should oppose the following EIPs from Istanbul:

EIP-1057: ProgPoW, a Programmatic Proof-of-Work
EIP-2029 and EIP-2031 State Rent`

Finally, I propose to consider adding:

ECIP-1043: Fixed DAG limit restriction
ECIP-1046: NiPoPoW*
ECIP-1047: Reduce gas limit
ECIP-1053: Add OpenRPC Service Discovery To JSON-RPC Services (unless these are already integrated)

*During this time we can debate whether to not do NiPoPoWs and do flyClient as per @zmitton.

soc1c:
I think it's too early to discuss timing and proposals. Let's focus on Atlantis and Agharta

sorpaas:
(Waiting on either #100 or #101 before getting this merged.)

PR #89, opened 2019-06-24
Create "assets" folder for images

TokenHash:
This is to create an "assets" folder on the ECIPs repo so images for ECIPs can be stored there.

This is described in the ECIP README.md file, but was not created:

`If your ECIP requires images, the image files should be included in a subdirectory of the `assets` folder for that ECIP as follows: assets/ecip-X. When linking to an image in the ECIP, use the related links such as ./assets/ecip-X/image.png.`


assets/hello
```diff
@@ -0,0 +1 @@
+ee
```
soc1c:
```suggestion
world
```

PR #92, opened 2019-06-27
Relocating "process" image into "assets" folder

TokenHash:
images go in the "assets" folder according to README.md instructions:

`If your ECIP requires images, the image files should be included in a subdirectory of the `assets` folder for that ECIP as follow: `assets/ecip-X`. When linking to an image in the ECIP, use the related links such as `./assets/ecip-X/image.png`.`


TokenHash:
Closing because image needs to be replaced.

PR #102, opened 2019-06-29
Lowercase filenames and add lang attribute to all specs

sorpaas:
* Lowercased spec filenames -- this fits better with the rest of folder structure and generated permalink is lowercased anyway.
* Add `lang` item to all specs.


index.html
```diff
@@ -13,20 +13,20 @@ <h2>ECIP Types</h2>
 
 <p>ECIPs are separated into a number of types, and each has its own list of ECIPs.</p>
 
-<h3>Standard Track ({{site.specs|where:"type","Standards Track"|size}})</h3>
+<h3>Standard Track ({{site.specs|where:"lang","en"|where:"type","Standards Track"|size}})</h3>
```
YazzyYaz: Cool stuff. This will make it easier to add more languages in the future for all of them and explore each ECIP.

sorpaas:
@YazzyYaz 

> What's the reason for lowercase filenames again?

It's nicer, and EIP repo uses lowercase as well. :)

PR #90, opened 2019-06-26
ECIP-69769: Submit Block RPC

BelfordZ:

PR #103, opened 2019-06-29
Re-open ECIP-69769: Submit Block RPC

sorpaas:
cc @BelfordZ. Originally PR #90. My mistake.


_specs/ecip-1062.md
```diff
@@ -0,0 +1,59 @@
...
+
+This removes the dependency for miners to have a geth node with accounts that need to be unlocked. Miners would be able to better isolate their private keys from the world, yeilding better security potential.
+
+### Implementation
+
+I think it's pretty straight forward. Looking at the above specification, we would be adding a eth_submitBlock rpc method. It's similar to eth_submitWork, except with fewer steps before broadcasting the block. 
+
+
+
+**NOTE: I am extremely open to suggestions for this. Please be liberal in your feedback. Thank you.**
```
sorpaas: @BelfordZ Maybe consider remove this line before the PR is merged.

_specs/ecip-1062.md
```diff
@@ -0,0 +1,59 @@
...
+}
+```
+
+### Rationale
+
+This removes the dependency for miners to have a geth node with accounts that need to be unlocked. Miners would be able to better isolate their private keys from the world, yeilding better security potential.
+
+### Implementation
+
+I think it's pretty straight forward. Looking at the above specification, we would be adding a eth_submitBlock rpc method. It's similar to eth_submitWork, except with fewer steps before broadcasting the block.
```
YazzyYaz: 1. Change eth_submitBlock to `eth_submitBlock`
2. Add "an" instead of "a" before `eth_submitBlock`.
2. Capitalize rpc to RPC
3. Change eth_submitWork to `eth_submitWork`

_specs/ecip-1062.md
YazzyYaz: Agreed.

_specs/ecip-1062.md
```diff
@@ -0,0 +1,59 @@
...
+      "result": {
+        "name": "isBlockAccepted",
+        "schema": {
+          "type": "boolean"
+        }
+      }
+    }
+  ]
+}
+```
```
YazzyYaz: Please link where relevant to Open-RPC when showing a JSON Schema for the method that uses Open-RPC. 

I understand Open-RPC is used in Multi-Geth, but the ECIP should take into consideration non-Open-RPC clients as well for clarity such as Parity and others.

Also, would be good to figure out if Parity can adopt at least for the case of the "classic" network.

_specs/ecip-1062.md
sorpaas: Addressed.

_specs/ecip-1062.md
YazzyYaz: I'm gonna delete the line if it's ok with you @sorpaas

sorpaas: Yeah!

YazzyYaz: @sorpaas fixed. Should be good to approve it now so we can merge.

BelfordZ: TY - been a busy weekend!

sorpaas: @BelfordZ please approve this PR if you feel it’s ready.

_specs/ecip-1062.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Rationale
+
+This removes the dependency for miners to have a geth node with accounts that need to be unlocked. Miners would be able to better isolate their private keys from the world, yeilding better security potential.
+
+### Implementation
+
+I think it's pretty straight forward. Looking at the above specification, we would be adding an `eth_submitBlock` RPC method. It's similar to `eth_submitWork`, except with fewer steps before broadcasting the block.
```
shanejonas: this section is usually reserved for links to the implementations in code if available.

PR #105, opened 2019-06-29
Add ./well-known/ecips location for all specs in this repo

sorpaas:
This exports all metadata of specs into a json file, which will be accessible from `https://ecips.ethereumclassic.org/.well-known/ecips/specs.json`, once this PR is merged.

`ecips` well-known location is registered in the [Well-Known URIs](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml) registry, and its definitions comply to [24-ECIPURI](https://ecips.that.world/24-ecipuri/).


.well-known/ecips/specs.json
```diff
@@ -0,0 +1,21 @@
+---
+layout: null
+---
+
+[
+    {% for page in site.specs %}
+    {
+        "uri": "ecips.ethereumclassic.org/{{ page.slug }}",
+        "documentUrl": "https://ecips.ethereumclassic.org{{ page.id }}",
```
phyro: is the `/` missing here before the page.id or is it added from the `_config.yml`?

sorpaas: No. The above is correct. `page.id` contains a leading `/`.

PR #112, opened 2019-07-22
ECIP-1063.md

26jesse:
What do you guys think of this idea? I like the idea of having decentralized oracles and smart contracts for Ethereum Classic.


26jesse:
ECIP: <1063>
Title: <Ethereum Classic Integrated Decentralized Oracles and Smart Contracts>
Author: <Jesse>
Discussions-To: <email address>
Status: <Draft>
Type: <Standards Track | Informational | Process>
Created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
Replaces: <ECIP number>
Superseded-By: <ECIP number>
Resolution: <url>
---
# Abstract
Imagine how we can use  ETC having oracle features which can assist Dapps and ERC-223 tokens to accomplish their goals. Although decentralized oracles rely on a source of information, it does not have to be from a centralized source. If an oracle is able to rely on multiple sources and is able to verify which information is the correct one, wouldn't it still work in a manner that is decentralized as possibly can? An example would be academia sources for a research paper, we have multiple sources that cannot be used for reliable information such as journals and memoirs.  However, with a decentralized oracle, it can then verify the reliability of an information and check if the sources have a track history of reliability such as from an academic institute. Information that is considered unreliable such as a  journal or memoir can be verified to be reliable if the journalist has a history of accuracy and reliability.  If the person who wrote the memoir is the primary source and has the evidence through other reliable sources that he is there, then the oracle can consider his memoir as a reliable source of information. Afterwards, then feed it to the smart contract to come with a reliable answer of what actually happened
# Motivation
To make better smart contracts even smarter than they are. The decentralized oracle can feed reliable information to the smart contract. A integrated decentralized oracle is useful for the Ethereum Classic network because relying on a third party for oracles means they can modify the code to suit the output. Integrated decentralized oracles create a tool that can better assist in problem solving beyond just academia, but has real life use cases that can be used for derivatives markets, gambling, accounting, and finance. An Ethereum Classic Decentralized oracle can be used not only for ETC projects, but can be used for other smart projects. Ethereum Classic's precedent as the first to integrate oracles and smart contracts means other smart contract platforms would look for an oracle solution that specifically address their issues
# Specification
## Decentralized Oracle Dapp
By becoming a Dapp, the decentralized Oracle can work independent of the network. This can be off-chain and help assist the smart contract
# Oracle Function as Part of ETC Network
Ethereum Classic can run the oracle within the limitations of its network and the smart contract. The oracle becomes just a tool of a smart contract to decide the information to input to the smart contract
## The Oracle as a Separate Layer or Separate Network
The oracle can act as a network, similar to the Lightening Network for Bitcoin, but instead the oracle network would act to make smart contracts more efficient and cheaper. The oracle network gwei rate would differ from the smart contract gwei rate
# Rationale
To further provide reliable information to the smart contract and to solve real world problems. This would create a greater decentralized network that provides reliable answers
# Implementation
All these examples of how we can use a decentralized oracle can be run with ETC gwei. Whether it will be a dapp or as a part of the ETC or as a separate network/layer, they will use ETC gwei power the oracle

1063
```diff
@@ -0,0 +1,26 @@
...
...racle features which can assist Dapps and ERC-223 tokens to accomplish their goals. Although decentralized oracles rely on a source of information, it does not have to be from a centralized source. If an oracle is able to rely on multiple sources and is able to verify which information is the correct one, wouldn't it still work in a manner that is decentralized as possibly can? An example would be academia sources for a research paper, we have multiple sources that cannot be used for reliable information such as journals and memoirs.  However, with a decentralized oracle, it can then verify the reliability of an information and check if the sources have a track history of reliability such as from an academic institute. Information that is considered unreliable such as a  journal or memoir can be verified to be reliable if the journalist has a history of accuracy and reliability.  If the person who wrote the memoir is the primary source and has the evidence through other reliable sources that he is there, then the oracle can consider his memoir as a reliable source of information. Afterwards, then feed it to the smart contract to come with a reliable answer of what actually happened
```
YazzyYaz: Please break down sentences into several lines. It's not user friendly to scroll sideways to read your content when reviewing it.

YazzyYaz:
Things left to do:
1) Add a space between each section as well. 
2) Fix the preamble by following our format. I give an example in a comment above, but here is it again as a reminder.
```
---
lang: en
ecip: 1044
title: Formalize IPFS hash into ECNS (Ethereum Classic Name Service) resolver
author: Phyrex Tsai <phyrex@portal.network>, Portal Network Team
status: Draft
type: Standards Track
category: ERC
created: 2018-05-12
---
```
3) Rename the file `1063` to `ecip-1063.md` and move it to the `_specs` folder.

PR #126, opened 2019-08-06
ECIP for Communicating with 3rd parties

bobthegrownup:
This proposal is a framework for 3rd party communications in decentralized systems


bobthegrownup:
I'm pretty sure i made a mistake and submitted to update ecip-x.md instead of create a new file.

PR #134, opened 2019-09-13
[ECIP-1054] Move Atlantis to Final, Add EIP-684

### LINKED ISSUE #67: ECIP-1054: Atlantis EVM and Protocol Upgrades
Atlantis Hardfork Meta
Enable the outstanding Ethereum Foundation Spurious Dragon and Byzantium network protocol upgrades on the Ethereum Classic network in a hard-fork code-named Atlantis to enable maximum compatibility across these networks.

etcatlantis_copy

To Discussion to : https://github.com/etclabscore/ECIPs/pull/10

soc1c:
Moves Atlantis to FINAL, closes #67 

Adds missing EIP-684, ref https://github.com/ethereum/EIPs/issues/2220 and https://github.com/ethereum/EIPs/pull/2252

### DIFF
--- a/_specs/ecip-1054.md
+++ b/_specs/ecip-1054.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1054
 title: Atlantis EVM and Protocol Upgrades
-status: Accepted
+status: Final
 type: Meta
 author: Isaac Ardis <isaac.a@etclabs.org>
 created: 2019-02-11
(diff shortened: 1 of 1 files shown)


GregTheGreek:
> EIP 684 applies retroactively starting from genesis. It should not has a hard fork number, and thus I don't believe it belongs here.

Where would it go? This is an ongoing debate on ETH - it should definitely be added to the Atlantis META. Similarly it should be added to the Byzantium META

sorpaas:
@GregTheGreek I still don't quite believe that 684 was the impact for the previous Classic Geth bug. The issue is that it shouldn't really affect anything before `CREATE2`. I think some information might have been mixed up in the process of that bug fix.

GregTheGreek:
I do agree, I don't think it was the root of the bug.
Nonetheless this still should be in the Atlantis meta, we don't need to follow the ETH eip process 1-for-1.

sorpaas:
@GregTheGreek If we agree that 684 wasn't the cause for any consensus bugs, then I believe my original argument stands. This is something that retroactively applies to genesis, and it does not fit into a Meta ECIP with a hard fork number. However, we should use a separate documentation to note that 684 is in effect.

sorpaas:
I created https://github.com/ethereumclassic/ECIPs/pull/142 to address the issue of 684.

PR #141, opened 2019-10-06
ECIP-1065: Kunming EVM and Protocol Upgrades

sorpaas:
This protocol upgrade provides a backward compatibility foundation to enable future protocol upgrades to be applied with significantly lower chance of breaking existing contracts.

See discussions section for details of impacts for the changes listed here.

### DIFF
--- a/_specs/ecip-1065.md
+++ b/_specs/ecip-1065.md
@@ -1,0 +1,83 @@
+---
+lang: en
+ecip: 1061
+title: Kunming EVM and Protocol Upgrades
+status: Draft
+type: Meta
+author: Wei Tang <hi@that.world>
+created: 2019-10-05
+---
+
+### Simple Summary
+
+This protocol upgrade provides a backward compatibility foundation to
+enable future protocol upgrades to be applied with significantly lower
+chance of breaking existing contracts.
+
+### Abstract
+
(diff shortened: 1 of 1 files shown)


zmitton:
The gas cost of contracts is often the cause of DOS vectors. The response to these vectors therefore *must* be to change *existing* gas costs. If existing gas costs are allowed to continue, DOS vector still exists.

In fact, I'm pretty sure all the DOS vectors so far discovered have been of this kind, fixed by changing gas costs. If this ECIP had been implemented, it would have had to have been overridden to change gas costs of already-deployed contracts.

realcodywburns:
The exploits usually are not long running contracts. The versioning could be retroactive for some number of blocks so that it breaks the contracts that are misbehaving but preserves longer term contracts.

sorpaas:
@zmitton @realcodywburns I want to restate the quite important thing -- account versioning is not for emergency hard forks. We could give it a try but that's not what this is designed for.

See "Emergency Hard Fork" in "Discussions" section of this document.

zmitton:
👍ok good youve got that wrapped up then. I still don't think we should do it unless Ethereum does it though.

sorpaas:
I changed the hard fork number to "undecided". Going to merge this if no objections.

PR #142, opened 2019-10-07
ECIP-1066: Active Ethereum Classic Network Description

sorpaas:
This creates an ECIP for the current active network description of Ethereum Classic mainnet. This will be the second ECIP document with the "active" status along with ECIP-1000.

### DIFF
--- a/_specs/ecip-1066.md
+++ b/_specs/ecip-1066.md
@@ -1,0 +1,24 @@
+---
+lang: en
+ecip: 1066
+title: Ethereum Classic Network Description
+status: Active
+type: Meta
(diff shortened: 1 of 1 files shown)


sorpaas:
@soc1c Do we have a hard fork at 2m?

soc1c:
The difficulty bomb was introduced at block 200k

sorpaas:
No. The only description I can find is geth's PR: https://github.com/ethereum/go-ethereum/pull/1588
Maybe we can add one later, given this indeed happened after the network launch.

BelfordZ:
nice

PR #145, opened 2019-10-18
ECIP-1055: Flyclient - Succinct PoW Using Merkle Mountain Ranges

TokenHash:
With author consent, transporting Flyclient ECLIP to ECIP in this repo.

Added "Flyclient" term in title as it is the common name for this kind of change.

### DIFF
--- a/_specs/ecip-1055.md
+++ b/_specs/ecip-1055.md
@@ -1,0 +1,141 @@
+---
+ecip: 1055
+title: Flyclient - Succinct PoW Using Merkle Mountain Ranges
+author: Zac Mitton @zmitton
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/144
+status: Draft
+type: Standards Track
+category (*only required for Standard Track): Core
+created: 2019-03-12
+replaces (*optional): 210, 1094, 641
+
+---
+
+## Simple Summary
+
+The addition of a Merkle-Mountain-Range (MMR) root to blockheaders
+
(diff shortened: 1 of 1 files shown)


_specs/ecip-1055.md
```diff
@@ -1,17 +1,15 @@
...
 type: Standards Track
-category (*only required for Standard Track): Core
+category: Core
 created: 2019-03-12
-replaces (*optional): 210, 1094, 641
-
 ---
 
-## Simple Summary
+## Simple Summaryhttps://github.com/etclabscore/ECIPs/blob/master/ECLIPs/ECIP-draft_succinct-proofs.md
```
mikeyb: ## [Simple Summary](https://github.com/etclabscore/ECIPs/blob/master/ECLIPs/ECIP-draft_succinct-proofs.md)

Should this be a url?

TokenHash: i don understand the question. i don't see that in the PR

do you mean to add that URL as a reference? if so, I don't think so because the ECIP is copied in its entirety

TokenHash: unless @zmitton want to reference it?

mikeyb: I was referring to this commit https://github.com/ethereumclassic/ECIPs/pull/145/commits/3b2f6b5c78f7c219d20f2febaa300c05fa7aa90b 

but it seems to have been reverted in the following commit.  ignore me

PR #149, opened 2019-10-21
Provided data points, risk considerations, implementation options, references.

Wolf-Linzhi:


_specs/ecip-1043.md
```diff
@@ -37,5 +94,13 @@ The following clients implemented ECIP
 - [] Parity
 - [] Mantis
 
+The following pools have indicated ECIP support
```
soc1c: The actual protocol governance should not be part of the specification document.

sorpaas: Let's remove this section.

PR #161, opened 2019-11-01
Rename Aztlán as Asgard

bobsummerwill:
Rename Aztlán as Asgard, which is a better known fictional place, easier to remember and has no ambiguity of spelling.

Twitter poll:
https://twitter.com/BobSummerwill/status/1188479134533484545

### DIFF
--- a/_specs/ecip-1061.md
+++ b/_specs/ecip-1061.md
@@ -1,7 +1,7 @@
 ---
 lang: en
 ecip: 1061
-title: Aztlán EVM and Protocol Upgrades
+title: Asgard EVM and Protocol Upgrades
 status: Draft
 type: Meta
 author: Talha Cross (@soc1c)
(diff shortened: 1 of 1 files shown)


bobsummerwill:
The Twitter poll is ecosystem signalling in support of this proposal.

The actually governance happens within the ECIP process, which is why I have created this pull request, for discussion within the ECIP process.

soc1c:
This Twitter poll indicates support for Aztlan.

https://twitter.com/BobSummerwill/status/1188070604286480384

drd34d:
A Twitter poll is definatelly not the place for **any** governance. What's next, a Straw Poll?

Asgard could be too similar to Agharta for the non-english speakers and thus slightly confusing. Don't see a reason to rename Aztlán/Aztlan at this point.

bobsummerwill:
@drd34d The poll was just that.  A poll.  Signalling.

The governance is happening on the #ecips channel in the ETC Discord (and in these comments), where incidentally we seem to have a rough consensus AGAINST any more "A-names".   Need something more different from Atlantis, and should also be easily understood by non-native speakers, yes.

YazzyYaz:
Given that @soc1c changed the spelling to Aztlan (which is the correct spelling for the Aztec realm and is the culturally appropriate way to call it), I see no reason to change the name to Asgard.

This is all bike shedding debates. Let's stick to Agharta and Aztlan and focus our naming efforts on future hardforks by having an open process for people to contribute their names.

bobsummerwill:
Fair enough, @YazzyYaz.  I think, as @stevanlohja highlighted, that:

Atlantis, Agharta and then Aztlán will be confusing for international audiences, but if I am outlier than I am happy to withdraw this proposal.

YazzyYaz:
@bobsummerwill I agree with you on the A lettering. A big part of the future hard forks would be to avoid the name A for a few hardforks post-Aztlan.

PR #170, opened 2019-11-06
ECIP-1040, ECIP-1068 and ECIP-1069: Account Versioning ECIPs and Extensions

sorpaas:
This adds the following specifications to ECIPs repo:

* ECIP-1040: The base layer of account versioning. The current rough consensus.
* ECIP-1068: Extension of ECIP-1040 to allow transaction to deploy multiple versions.
* ECIP-1069: Extension of ECIP-1040 to allow CREATE/CREATE2 to deploy multiple versions.

### DIFF
--- a/_specs/ecip-1040.md
+++ b/_specs/ecip-1040.md
@@ -8,6 +8,7 @@ status: Draft
 type: Standards Track
 category: Core
 created: 2017-12-30
+license: Apache-2.0
 ---
 
 ## Simple Summary
--- a/_specs/ecip-1040.md
+++ b/_specs/ecip-1040.md
@@ -1,0 +1,238 @@
+---
+lang: en
+ecip: 1040
+title: Generalized Account Versioning Scheme
+author: Wei Tang (@sorpaas)
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/169
+status: Draft
+type: Standards Track
+category: Core
+created: 2017-12-30
+---
+
+## Simple Summary
+
+Introduce account versioning for smart contracts so upgrading the VM
+or introducing new VMs can be easier.
+
+## Abstract
+
+This defines a method of hard forking while maintaining the exact
+functionality of existing account by allowing multiple versions of the
+virtual machines to execute in the same block. This is also useful to
+define future account state structures when we introduce the on-chain
+WebAssembly virtual machine.
+
+## Motivation
+
+By allowing account versioning, we can execute different virtual
+machine for contracts created at different times. This allows breaking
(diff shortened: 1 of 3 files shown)


bobsummerwill:
Awesome!  It will be great to start progressing these within the ECIP process :-)

The licensing should be specified in the metadata for each of these files, shouldn't it?
Not very as text at the bottom.

sorpaas:
@bobsummerwill Fixed.

_specs/ecip-1040.md
```diff
@@ -0,0 +1,239 @@
+---
+lang: en
+ecip: 1040
+title: Generalized Account Versioning Scheme
+author: Wei Tang (@sorpaas)
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/169
+status: Draft
+type: Standards Track
+category: Core
+created: 2017-12-30
```
YazzyYaz: Should this be changed to today's date instead or when you created the specific document as an ECIP with Apache License?

_specs/ecip-1068.md
```diff
@@ -0,0 +1,72 @@
+---
+lang: en
+ecip: 1068
+title: Account Versioning Extension for Contract Creation Transaction
+author: Wei Tang (@sorpaas)
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/169
+status: Draft
+type: Standards Track
+category: Core
+created: 2019-06-24
```
YazzyYaz: Since this is also being submitted as an ECIP with an Apache License, would it make sense to have created reflect the date of the submission of the ECIP itself?

_specs/ecip-1069.md
```diff
@@ -0,0 +1,78 @@
+---
+lang: en
+ecip: 1069
+title: Account Versioning Extension for CREATE and CREATE2
+author: Wei Tang (@sorpaas)
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/169
+status: Draft
+type: Standards Track
+category: Core
+created: 2019-06-24
```
YazzyYaz: Same point here as others regarding date of submission of an ECIP specifically vs the original EIP. Because you're adding the Apache License as part of the submission, I think adding the correct date for the PR submission makes sense.

_specs/ecip-1040.md
sorpaas: I don’t see why that’s needed. There’s technically no EIPs for those two extensions as they are never merged.

YazzyYaz: Alrighty, just wanted to point it out. Resolving.

PR #179, opened 2019-11-08
ECIP-1071: Methods to Merge Ethereum Blockchains Together

sorpaas:


bobsummerwill:
Is this missing a catagory?
Should be ECBP?
Does this even need to be an ECIP?
Can just remain blog post / ETC Summit talk?

phyro:
@bobsummerwill it has a category `core` specified. It also seems like an implementation kind of thing so I believe core is appropriate, no?

Issue #182, opened 2019-11-13
Reject "ASIC resistance" myth


bobsummerwill:
Discussion related to https://github.com/ethereumclassic/ECIPs/pull/183.

**"ASIC resistance" is a myth.  The ETC ecosystem rejects this myth.**

drd34d:
I think you got the wrong url. You're probably looking for "Twitter.com"

PR #183, opened 2019-11-13
Add ECIP for Reject "ASIC resistance" myth

bobsummerwill:
Informational ECIP:

**"ASIC resistance" is a myth.  The ETC ecosystem rejects this myth.**

Discussion to https://github.com/ethereumclassic/ECIPs/issues/182, please!

### DIFF
--- a/_specs/ecip-asic-resistance.md
+++ b/_specs/ecip-asic-resistance.md
@@ -14,6 +14,93 @@ discussions-to: https://github.com/ethereumclassic/ECIPs/issues/182
 
 "ASIC resistance" is a myth.  The ETC ecosystem rejects this myth.
 
+This ECIP seeks to bring certainty that decisions around ETC mining
+are not being guided by an easily disprovable myth, but by the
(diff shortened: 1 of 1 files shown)


bobsummerwill:
Discussion to https://github.com/ethereumclassic/ECIPs/issues/182, please!

bobsummerwill:
I will create a longer Rationale section in here based on prior discussions, on ETC Summit content from the individual talks and from the mining panel, and based on Discord / Twitter / WeChat / other debate around this ECIP, as that takes shape.

zmitton:
Mostly agree. Not sure myth is the right word. It's a game of whack-a-mole. I think it is the constant "configuring" required to make it work which ETC should reject. It puts too much power to the teams and gives a given software implementation a limited lifespan (before it needs upgrade to avoid ASICs)

bobsummerwill:
Obsoleted by rejection of ProgPOW.   I will just turn this into a blog post instead.

https://github.com/ethereumclassic/ECIPs/pull/193

PR #184, opened 2019-11-14
ECIP-1072, ECIP-1073, ECIP-1074: Backward Compatible Options for Istanbul Hard Forks

sorpaas:
This adds three new hard fork meta, with different options to apply the Istanbul upgrade in a backward compatible way:

* ECIP-1072: Apply Istanbul directly, but remove the offending EIP-1884.
* ECIP-1073: Apply Istanbul with account versioning ECIP-1040 / EIP-1702. Currently both Parity and Geth have the implementation required. However, the drawback, as discussed last time, is that this will require a new account version for every hard fork.
* ECIP-1074: Apply Istanbul with account versioning, and with forward-compatible EVM changes. Currently no client has implemented 39-UNGAS and 40-UNUSED yet, so this will require more work, and we will need to maintain our own ecosystem tools. However, it gives us a good future foundation for modifying EVMs further. This is basically combination of ECIP-1065 Kunming hard fork with Istanbul.

### DIFF
--- a/_specs/ecip-1072.md
+++ b/_specs/ecip-1072.md
@@ -7,6 +7,7 @@ type: Meta
 author: Wei Tang (@sorpaas)
 created: 2019-11-14
 discussions-to: https://github.com/ethereumclassic/ECIPs/issues/81
+license: Apache-2.0
 ---
 
 ### Simple Summary
--- a/_specs/ecip-1072.md
+++ b/_specs/ecip-1072.md
@@ -1,0 +1,79 @@
+---
+lang: en
+ecip: 1072
+title: Yingchun EVM and Protocol Upgrades
+status: Draft
+type: Meta
+author: Wei Tang (@sorpaas)
+created: 2019-11-14
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/81
+---
+
+### Simple Summary
+
+Enable the outstanding Ethereum Foundation _Istanbul_ network protocol
+upgrades on the Ethereum Classic network in a hard-fork code-named
+_Yingchun_ to enable maximum compatibility across these networks.
+
+In particular, delay EIP-1884 activation until backward compatibility
+solutions are in place on Ethereum Classic.
+
+### Abstract
+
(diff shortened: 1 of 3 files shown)


_specs/ecip-1074.md
```diff
@@ -0,0 +1,96 @@
...
+author: Wei Tang (@sorpaas)
+created: 2019-11-14
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/81
+---
+
+### Simple Summary
+
+Enable the outstanding Ethereum Foundation _Istanbul_ network protocol
+upgrades on the Ethereum Classic network in a hard-fork code-named
+_Xichun_ to enable maximum compatibility across these networks.
```
YazzyYaz: This needs be renamed to `Tanchun`

_specs/ecip-1072.md
```diff
@@ -0,0 +1,79 @@
...
+
+- [EIP-152](https://eips.ethereum.org/EIPS/eip-152): Add Blake2
+  compression function `F` precompile
+- [EIP-1108](https://eips.ethereum.org/EIPS/eip-1108): Reduce
+  alt_bn128 precompile gas costs
+- [EIP-1344](https://eips.ethereum.org/EIPS/eip-1344): Add ChainID
+  opcode
+- [EIP-2028](https://eips.ethereum.org/EIPS/eip-2028): Calldata gas
+  cost reduction
+- [EIP-2200](https://eips.ethereum.org/EIPS/eip-2200): Rebalance
```
YazzyYaz: Those EIPs are the same as added ECIP-1061, no? I see the differences in the ECIP-1073 and ECIP-1074, but I think, unless I'm mistaken, Yingchun and Aztlan are the same essentially, right?

sorpaas: @YazzyYaz Yingchun is Aztlan with EIP-1884 stripped out!

YazzyYaz: Ah I see it, sorry my bad.

PR #185, opened 2019-11-15
Rename Aztlan ECIPs with "edition format"

sorpaas:
Rename so that Aztlan ECIPs clearly shows four different options:

- ECIP-1061 as ECIP-1061A
- ECIP-1072 as ECIP-1061B
- ECIP-1073 as ECIP-1061C
- ECIP-1074 as ECIP-1061D

Note that we cannot yet merge this because names like "ECIP-1061A" does not comply to the ECIP-1000 procedure. It's a trivial change, but I'll wait @bobsummerwill to know whether we can just do that (or should go through the whole process of creating another ECIP).

### DIFF
--- a/_specs/ecip-1061a.md
+++ b/_specs/ecip-1061a.md
@@ -1,7 +1,7 @@
 ---
 lang: en
-ecip: 1061
-title: Aztlán EVM and Protocol Upgrades
+ecip: 1061a
+title: Aztlán EVM and Protocol Upgrades (Yuanchun Edition)
 status: Draft
 type: Meta
 author: Talha Cross (@soc1c)
--- a/_specs/ecip-1061b.md
+++ b/_specs/ecip-1061b.md
@@ -1,7 +1,7 @@
 ---
 lang: en
-ecip: 1074
-title: Tanchun EVM and Protocol Upgrades
+ecip: 1061b
+title: Aztlán EVM and Protocol Upgrades (Yingchun Edition)
 status: Draft
 type: Meta
 author: Wei Tang (@sorpaas)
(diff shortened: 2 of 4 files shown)


bobsummerwill:
I don't think A/B/C/D is needed.
Just stick with the numbers, IMHO.
Easy to make a group ECIP for them all.

Maybe Afri's existing Aztlán one is that, and we just nest Sorpaas three proposals under that, and make a new ECIP for the plan currently in that?

bobsummerwill:
So 1061 becomes a "options for next HF", referencing 1072, 1073, 1074  plus a new ECIP which is "the plan" which Afri had?

sorpaas:
@bobsummerwill That will mean putting 1061 in a situation other than "Draft", because it is somewhat "official" in that it's the "root" ECIP for the next hard fork. It's a viable option but we'd still need a new ECIP status for that.

bobsummerwill:
I think that 1061 would still be "Draft" for the time being, and so would the four "options".

Then maybe 1061 does to "Accepted" when we decide on a single plan?   Or maybe just gets thrown away at that point, with the ECIP for the plan which we are doing taking over that role.

bobsummerwill:
Something I am thinking here, which is maybe not clear to others, is that having contradictory plans within the ECIP process is fine.   Just fine.   Maybe for years even.   Multiple plans, all being worked on.

You only need to resolve that stuff when it comes to decision points on what goes into a hard-fork.  We are not at that decision point for the March fork yet.

sorpaas:
@bobsummerwill Another point I have is that this is really just about ECIP numbers. "ECIP-1074" is really functionally equivalent to "ECIP-1061D". But with numbers such as "1061A/B/C/D", we make it much easier for everyone to recognize that those are different hard fork proposals that are supposed to happen at the same time.

bobsummerwill:
@sorpaas I don't think that A/B/C/D really helps, to be honest.
I would just stick with simple numbers, rather than inventing a new scheme.

It will be obvious from looking at the ECIPs that they are alternatives.   Those people who are participating in the ECIP process close enough to know specific numbers don't need the "hint", I don't think.

No need to complicate things, IMHO.

sorpaas:
@bobsummerwill Okay. Then I think this will not move forward. Let's keep things as it is currently.

Issue #174, opened 2019-11-07
Core Devs Call: Mining Algorithm Upgrade


soc1c:
### ETC Core Devs Call - Mining Algorithm Upgrade

When: **Thursday, November 21, 2019, 1pm UTC**, 60 minutes max.

Where: **Ethereum Classic Discord** https://discord.gg/dwxb6nf `#ecips` channel. Will use/create a voice channel _ad hoc._

### Agenda

- Quick client teams check-in
  - Parity Ethereum / Parity Tech
  - Geth Classic / ChainSafe, ETC Labs Core
  - Multi-Geth / Multi-Geth, ETCLabs Core 
  - Hyperledger Besu / ChainSafe, PegaSys
- Defining a process
  - keeping the call technical?
  - ECBP-1076 "algo change process" ref #187
- Discuss options for mining algorithms
  - ECIP-1049 "SHA-3" @p3c-bot ref #8 #13
  - ECIP-1070 "ProgPOW" @YazzyYaz ref #171 #172 
  - Status Quo: "Ethash" ...
- Agree on a joint way forward on the three proposals above
- Commit to a timeline
- Focus on a testnet
- Discuss anything else related to Mining
- Bonus: 
  - ECIP-1043 "DAG Size Limit" @realcodywburns @Wolf-Linzhi ref #11 #138
  - ECIP-1047 "Block Gas Limit" @pyskell @zmitton ref #6 #14

Please comment to add items to the agenda

bobsummerwill:
We need to discuss gaslimit too:

@pyskell has submitted this proposal - https://ecips.ethereumclassic.org/ECIPs/ecip-1047.
@zmitton has indicated he has an alternative
And I will create an alternative ECIP as well.

I would argue that gaslimit and DAG are more important to tackle first, and algorithm options second.

The gaslimit and DAG are immediate issues with the current setup.

TokenHash:
"Agree on a joint way forward on the **four** proposals above"

soc1c:
Updated agenda

bobsummerwill:
Yes, @TokenHash.

I would suggest:

BLOCK 1:
- Status quo on gas, or impose gas limit
- Status quo on DAG, or introduce DAG size limit

BLOCK 2:
- Status quo on Ethash, or SHA3 or ProgPOW

[...]

PR #187, opened 2019-11-16
ECBP-1076: Mining algorithm change process

soc1c:
This meta document is agnostic to any mining algorithm. Its sole purpose is specifying a process how to openly determine and select a mining algorithm for Ethereum Classic.

ref #174

### DIFF
--- a/_specs/ecip-10746.md
+++ b/_specs/ecip-10746.md
@@ -83,6 +83,11 @@ A Multi-Geth node configured to vote for Ethash with DAG size limit (ECBP-1076-B
 ```
 geth --miner.extradata "76B01 Funny Classic Pool" --miner.gaslimit 1000000 [OPTIONS]
 ```
+A Besu node configured to vote for ProgPoW (ECBP-1076-D) and an increase of the block gas limit:
+
+```
+besu --miner-extra-data="76D15 LOL-lonely Miner 1337" --target-gas-limit=15000000 [OPTIONS]
+```
 
 ### Implementation
 All clients implement custom extra data and block gas limit target voting. No custom implementation is required.
--- a/_specs/ecip-10746.md
+++ b/_specs/ecip-10746.md
@@ -1,0 +1,91 @@
+---
+ecip: 1076
+title: Mining Algorithm Change Process
+author: Talha Cross (@soc1c)
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/174
+status: Draft
+type: ECBP
+created: 2019-11-16
+---
+
+### Abstract
+This meta document is agnostic to any mining algorithm. Its sole purpose is specifying a process how to openly determine and select a mining algorithm for Ethereum Classic.
+
+### Motivation
+The decision to change the Ethereum Classic mining algorithm - or to keep it as is - is not an easy one to make. It should be done by broad consensus with all stakeholders involved - miners, investors, application and core developers, and anyone else who feels having a stake in ETC. This meta document proposes a process of how to facilitate the potential change of the mining algorithm and parameters.
+
+### Available Options
+Currently, the following options are debatable.
+
+- **ECBP-1076-A**: _"Ethash Status Quo."_ The mining algorithm will be untouched as it is since genesis, as **Ethash** function with a statically **increasing DAG**.
+- **ECBP-1076-B**: _"Ethash Limited DAG."_ The mining algorithm will be untouched as it is since genesis, as **Ethash** function. However, the **DAG size will be limited** as specified in [ECIP-1043](https://ecips.ethereumclassic.org/ECIPs/ecip-1043).
+- **ECBP-1076-C**: _"SHA3."_ The mining algorithm will be **changed to SHA3** as specified in [ECIP-1049](https://ecips.ethereumclassic.org/ECIPs/ecip-1049).
+- **ECBP-1076-D**: _"ProgPoW."_ The mining algorithm will be **changed to ProgPow** as specified in [ECIP-1070](https://ecips.ethereumclassic.org/ECIPs/ecip-1070).
+
+Eventually, the community has to decide on one.
+
+### Process
+The following process facilitates all stakeholders in various stages.
+
+1. _"Tech Stage."_ Core and application developers meet in regular calls to evaluate the feasibility of the Ethash status quo, the limited DAG size proposal (ECIP-1043), the SHA3 proposal (ECIP-1049), and the ProgPoW proposal (ECIP-1070).
+
+    If there is no technical objection to each of the proposals, the proposals can be **moved to "Last Call"** state, regardless of whether they will be considered in future or not. As a result, none, one, or many competing proposals can be in "Last Call" state at the same time.
+
+    This stage is finished once all proposals are either in "Last Call," "Withdrawn," or "Rejected" state.
+
+2. _"Miner Stage."_ Solo miners, mining farms, and mining pools can signal support by adjusting their mining node to signal in favor of one of the proposals that was promoted to "Last Call" state in the previous stage. Details on the signaling can be found in the subsequent section.
+
+    This stage is finished once more than **70% of the last 100_000 blocks** were signed with a signal in favor of one proposal.
+3. _"Decision Stage."_ The proposal that passed the technical stage to "Last Call" state and received 70% approval by the miner community, **shall be considered as "Accepted,"** all other competing proposals shall be considered as "Rejected." The EIP Editors are responsible to facilitate this status update to the proposals according to this process definition.
+
+    The block number where the 70% threshold is breached the first time is called the `threshold_blocknumber`. It automatically defines the `activation_blocknumber` which is:
+
+    ```
+    activation_blocknumber = threshold_blocknumber + 1_000_000
+    ```
+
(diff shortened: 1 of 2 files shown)


bobsummerwill:
My primary objection to this process as stated is that it puts the decision-making on change of mining algorithm purely in the hands of the miners.

They are not the only stakeholders and should not be the ones making that decision.

Miners will vote in their own interests, which do not necessarily coincide with the needs of the network.  Gaslimit voting for ETH1 is a primary example of this, where gaslimit has been voted beyond the limits of the client software to cope - resulting is massive centralization.

There is no counter-balance there, and there is no counter-balance here.

soc1c:
this is a draft. please add as many steps to the process as you wish

phyro:
I agree with @bobsummerwill . This kind of voting/signaling has the same incentive flaw as the DAO carbon vote. Miners are incentivized to optimize for themselves in a similar way that the DAO token holders were incentivized to bail themselves out.
The nice thing about this is that it is automated though.

soc1c:
I will champion this proposal unless you come up with a better idea. And I am open for suggestions or adding additional signals.

Nobody wants to spend tens of hours on calls.

_specs/ecip-10746.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Motivation
+The decision to change the Ethereum Classic mining algorithm - or to keep it as is - is not an easy one to make. It should be done by broad consensus with all stakeholders involved - miners, investors, application and core developers, and anyone else who feels having a stake in ETC. This meta document proposes a process of how to facilitate the potential change of the mining algorithm and parameters.
+
+### Available Options
+Currently, the following options are debatable.
+
+- **ECBP-1076-A**: _"Ethash Status Quo."_ The mining algorithm will be untouched as it is since genesis, as **Ethash** function with a statically **increasing DAG**.
+- **ECBP-1076-B**: _"Ethash Limited DAG."_ The mining algorithm will be untouched as it is since genesis, as **Ethash** function. However, the **DAG size will be limited** as specified in [ECIP-1043](https://ecips.ethereumclassic.org/ECIPs/ecip-1043).
+- **ECBP-1076-C**: _"SHA3."_ The mining algorithm will be **changed to SHA3** as specified in [ECIP-1049](https://ecips.ethereumclassic.org/ECIPs/ecip-1049).
```
YazzyYaz: I'd change it to Keccak256 for consistency and more technical accuracy and to reflect the language in ECIP-1049

soc1c: Keccak256 might be confused with the Keccak256 candidate used in Ethereum though.

YazzyYaz: Perhaps then Keccak256 (SHA3)? Maybe @prestwich can weigh in with his thoughts too? Its only to ensure the language communicates the correct engineering implementation.

sorpaas:
ECBP-1076 is basically MINERVOTE. But TBH, if we're going to the route of letting miners decide PoW algorithm change, then I'd suggest we do something more sophisticated, such as 27-MINERVOTE, rather than ad hoc process that still involves human factors. https://specs.that.world/27-minervote/

If you are a miner, I would guess the most important thing you want to review are:

* Voting range: in ECBP-1076 this is unspecified (?). In 27-MINERVOTE this is a known recurrence period, usually several weeks to a month.
* Activation time: in ECBP-1076 this is 6 months (!). In 27-MINERVOTE this is again, usually several weeks to a month.
* Threshold: in ECBP-1076 this is 70%, meaning there may be still 30% of the miners against a particular algorithm change. In 27-MINERVOTE we don't have a fixed number, but for things controversial like this we may want a higher threshold, such as 90%.

soc1c:
Thanks for the feedback.

- The voting range is indefinite. Any block that reachs the threshold should be considered as threshold block. Happy to talk about a sensible range though, I just cannot imagine any, yet.
- It's six months because the change is substantial and we need all clients to implement it. You cannot switch to Keccak256 in a couple of weeks.
- It makes sense to bump this to 90% if it appears to be controversial. However, this would mean that Ethash ASIC producers only need 10% of the ETC hashrate to ensure a change will never happen. It might be sensible to go with 70%.

_specs/ecip-10746.md
prestwich: because Ethereum introduced a lot of naming confusion, you might consider defining this by the output it produces for a set of inputs. 

e.g. the hash function that produces the digest `0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470` when given the preimage `0x`,  which we refer to as `keccak256`

sorpaas: Yeah agree with @YazzyYaz here -- `keccak256` would be more precise.

soc1c: Done

_specs/ecip-1076.md
```diff
@@ -0,0 +1,97 @@
...
+status: Draft
+type: ECBP
+created: 2019-11-16
+---
+
+### Abstract
+This meta document is agnostic to any mining algorithm. Its sole purpose is specifying a process how to openly determine and select a mining algorithm for Ethereum Classic.
+
+### Motivation
+The decision to change the Ethereum Classic mining algorithm - or to keep it as is - is not an easy one to make. It should be done by broad consensus with all stakeholders involved - miners, investors, application and core developers, and anyone else who feels having a stake in ETC. This meta document proposes a process of how to facilitate the potential change of the mining algorithm and parameters.
```
mikeyb: > and anyone else who feels having a stake in ETC
Should this be?
> and anyone else who feels they have a stake in ETC

_specs/ecip-1076.md
```diff
@@ -0,0 +1,97 @@
...
+    - `B` (`0x42`): ECBP-1076-B
+    - `C` (`0x43`): ECBP-1076-C
+    - `D` (`0x44`): ECBP-1076-D
+3. Two numeric bytes that vote on the default gas limit as specified in [ECIP-1047](https://ecips.ethereumclassic.org/ECIPs/ecip-1047).
+    - `00` (`0x30 0x30`): abstain / don't care _(optional)_
+    - `01` (`0x30 0x31`): support ECIP-1047 client defaults with 1 million gas block gas limit
+    - `08` (`0x30 0x38`): reject ECIP-1047 client defaults, explicitly supporting the 8 million gas block gas limit defaults
+    - `99` (`0x39 0x39`): any other number between `00` and `99` suggests a competing default block gas limit
+
+In addition to the numeric vote on the gas limit with the extra data field, mining nodes are encouraged to utilized block gas target limit voting with the suitable configuration flags for their mining nodes.
```
mikeyb: > are encouraged to utilized
Should be ?
> are encouraged to utilize

_specs/ecip-1076.md
```diff
@@ -0,0 +1,97 @@
...
+
+- **ECBP-1076-A**: _"Ethash Status Quo."_ The mining algorithm will be untouched as it is since genesis, as **Ethash** function with a statically **increasing DAG**.
+- **ECBP-1076-B**: _"Ethash Limited DAG."_ The mining algorithm will be untouched as it is since genesis, as **Ethash** function. However, the **DAG size will be limited** as specified in [ECIP-1043](https://ecips.ethereumclassic.org/ECIPs/ecip-1043).
+- **ECBP-1076-C**: _"Keccak256 (SHA3)."_ The mining algorithm will be **changed to Keccak256** as specified in [ECIP-1049](https://ecips.ethereumclassic.org/ECIPs/ecip-1049).
+- **ECBP-1076-D**: _"ProgPoW."_ The mining algorithm will be **changed to ProgPow** as specified in [ECIP-1070](https://ecips.ethereumclassic.org/ECIPs/ecip-1070).
+
+Eventually, the community has to decide on one.
+
+### Process
+The following process facilitates all stakeholders in various stages.
```
zmitton:
```suggestion
The following process facilitates all stakeholders in various stages, with "Tech Stage" and the "Miner Stage" to occur in parallel:
```

_specs/ecip-1076.md
```diff
@@ -0,0 +1,97 @@
...
+
+1. _"Tech Stage."_ Core and application developers meet in regular calls to evaluate the feasibility of the Ethash status quo, the limited DAG size proposal (ECIP-1043), the Keccak256 proposal (ECIP-1049), and the ProgPoW proposal (ECIP-1070).
+
+    If there is no technical objection to each of the proposals, the proposals can be **moved to "Last Call"** state, regardless of whether they will be considered in future or not. As a result, none, one, or many competing proposals can be in "Last Call" state at the same time.
+
+    This stage is finished once all proposals are either in "Last Call," "Withdrawn," or "Rejected" state.
+
+2. _"Miner Stage."_ Solo miners, mining farms, and mining pools can signal support by adjusting their mining node to signal in favor of one of the proposals that was promoted to "Last Call" state in the previous stage. Details on the signaling can be found in the subsequent section.
```
zmitton:
```suggestion
2. _"Miner Stage."_ Solo miners, mining farms, and mining pools can signal support by adjusting their mining node to signal in favor of one of the proposals specified here. Details on the signaling can be found in the subsequent section.
```

soc1c:
@zmitton can you review #194 instead?

PR #192, opened 2019-11-21
Inline specification for 39-GAS and 40-UNUSED from that.world.

bobsummerwill:
Better to have these self-contained specifications right here.
They are not complex.

### DIFF
--- a/_specs/ecip-1065.md
+++ b/_specs/ecip-1065.md
@@ -28,11 +28,22 @@ At hard fork block, enable [EIP-1702](https://eips.ethereum.org/EIPS/eip-1702):
 * Define the previous EVM version as version `0`.
 * Define a new version `1`, with the following EVM modifications
   applied on it.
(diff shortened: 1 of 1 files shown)


sorpaas:
@bobsummerwill The direction is fine for me. But I'll just need to do some tidy up work on the actual specifications before merging, as the thing presented in this PR is, IMO, unreadable.

On a side note, while 39-UNGAS itself currently looks not complex, it may actually require some refactorings in EVM, not yet sure the amount of work. I'll explain in more details about what I mean in this case. Specifically, there're some corner cases that must be clearly specified.

bobsummerwill:
If you want to redo something equivalent, but with better wording yourself, go for it, and I will close this.

bobsummerwill:
I am going to close this, @sorpaas.

You know what I meant with this direction, but it is better for you to do the work.

PR #193, opened 2019-11-21
Update ECIP-1070 Status to Rejected

YazzyYaz:
As per the Mining Call and Issue Link Summary: https://github.com/ethereumclassic/ECIPs/issues/174#issuecomment-557106181

Moving ECIP-1070: ProgPoW on ETC to Rejected Status with the agreement that someone who wants to champion this proposal can do so on a new ECIP in the future.


realcodywburns:
Ack, rejected due to lack of support.

PR #194, opened 2019-11-21
ecbp 1076: activate miner signaling process

soc1c:
activating the ecbp as per call today https://github.com/ethereumclassic/ECIPs/issues/174#issuecomment-557106181

### DIFF
--- a/_specs/ecip-1076.md
+++ b/_specs/ecip-1076.md
@@ -1,19 +1,19 @@
 ---
 lang: en
 ecip: 1076
-title: Mining Algorithm Change Process
+title: Mining Algorithm Signaling Process
 author: Talha Cross (@soc1c)
 discussions-to: https://github.com/ethereumclassic/ECIPs/issues/174
-status: Draft
+status: Active
 type: ECBP
 created: 2019-11-16
 ---
 
 ### Abstract
-This meta document is agnostic to any mining algorithm. Its sole purpose is specifying a process how to openly determine and select a mining algorithm for Ethereum Classic.
+This meta document is agnostic to any mining algorithm. Its sole purpose is specifying a process for miners to signal their preferred mining algorithm for Ethereum Classic.
 
 ### Motivation
-The decision to change the Ethereum Classic mining algorithm - or to keep it as is - is not an easy one to make. It should be done by broad consensus with all stakeholders involved - miners, investors, application and core developers, and anyone else who feels they have a stake in ETC. This meta document proposes a process of how to facilitate the potential change of the mining algorithm and parameters.
+The decision to change the Ethereum Classic mining algorithm - or to keep it as is - is not an easy one to make. It should be done by broad consensus with all stakeholders involved - miners, investors, application and core developers, and anyone else who feels they have a stake in ETC. This meta document proposes a process of how miners could signal in favor of one potential change of the mining algorithm and parameters.
 
 ### Available Options
 Currently, the following options are debatable.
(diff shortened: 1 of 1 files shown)


_specs/ecip-1076.md
```diff
@@ -1,58 +1,39 @@
 ---
 lang: en
 ecip: 1076
-title: Mining Algorithm Change Process
+title: Mining Algorithm Signaling Process
 author: Talha Cross (@soc1c)
 discussions-to: https://github.com/ethereumclassic/ECIPs/issues/174
-status: Draft
+status: Active
```
sorpaas: This is invalid. Signaling is a hard fork process, and thus this should follow "Last Call" -> "Accepted" -> "Final" route, instead of "Active".

soc1c: It's not a hard fork process. It's non-binding signaling.

sorpaas: @soc1c Still, there's no justification to move this to "Active" status. I believe that's just wrong.

soc1c: Quote ECIP-1000:

> _"Process ECIPs may also be moved to a status of “Active” instead"_

This clearly is a process ECIP.

sorpaas: @soc1c I would put this in the same category as specifications like ERC-20. I don't believe they can be moved to "Active" status, thus the same for this ECBP.

soc1c:
please allow editors and community members to review this change

tzdybal:
1. ECBP is a category of "standard track" ECIPs.
2. As per ECIP-1000 you cannot move ECIP/ECBP from `Draft` to `Active`. `Active` can be used only instead of `Final`. And there is also `Last call` and `Accepted` state between `Draft` and `Final`.
    > Some Informational ECIPs, which are considered process ECIPs, may also be moved to a status of `Active` instead of `Final` if they are never meant to be completed, e.g. this ECIP-1000.
3. I agree that `Active` is not a proper state for this ECBP.

_specs/ecip-1076.md
```diff
@@ -1,58 +1,39 @@
...
+### Signaling Process
+Solo miners, mining farms, and mining pools can signal support by adjusting their mining node to signal in favor of one of the proposals that was promoted to "Last Call" state in the previous stage. Details on the signaling can be found in the subsequent subsections.
 
-    If there is no technical objection to each of the proposals, the proposals can be **moved to "Last Call"** state, regardless of whether they will be considered in future or not. As a result, none, one, or many competing proposals can be in "Last Call" state at the same time.
+The signaling process shall be considered finished once more than **70% of the last 1_000_000 blocks** were signed with a signal in favor of one proposal.
 
-    This stage is finished once all proposals are either in "Last Call," "Withdrawn," or "Rejected" state.
-
-2. _"Miner Stage."_ Solo miners, mining farms, and mining pools can signal support by adjusting their mining node to signal in favor of one of the proposals that was promoted to "Last Call" state in the previous stage. Details on the signaling can be found in the subsequent section.
```
zmitton:
```suggestion
2. _"Miner Stage."_ Solo miners, mining farms, and mining pools can signal support by adjusting their mining node to signal in favor of one of the proposals specified here. Details on the signaling can be found in the subsequent section.
```

_specs/ecip-1076.md
```diff
@@ -1,58 +1,39 @@
...
 - **ECBP-1076-B**: _"Ethash Limited DAG."_ The mining algorithm will be untouched as it is since genesis, as **Ethash** function. However, the **DAG size will be limited** as specified in [ECIP-1043](https://ecips.ethereumclassic.org/ECIPs/ecip-1043).
 - **ECBP-1076-C**: _"Keccak256 (SHA3)."_ The mining algorithm will be **changed to Keccak256** as specified in [ECIP-1049](https://ecips.ethereumclassic.org/ECIPs/ecip-1049).
-- **ECBP-1076-D**: _"ProgPoW."_ The mining algorithm will be **changed to ProgPow** as specified in [ECIP-1070](https://ecips.ethereumclassic.org/ECIPs/ecip-1070).
+- ~~**ECBP-1076-D**: _"ProgPoW."_ The mining algorithm will be **changed to ProgPow** as specified in [ECIP-1070](https://ecips.ethereumclassic.org/ECIPs/ecip-1070).~~ 
 
-Eventually, the community has to decide on one.
+    _Note: ProgPoW was rejected by the community on Nov/20, 2019. However, miners can still signal for this option if they prefer this over the other options._
 
-### Process
-The following process facilitates all stakeholders in various stages.
```
zmitton:
```suggestion
The following process facilitates all stakeholders in various stages, with "Tech Stage" and the "Miner Stage" to occur in parallel:
```

PR #197, opened 2019-11-27
Enable 44-VERTXN extension to make ecosystem tooling backward compatible

sorpaas:
This enables 44-VERTXN extension for EIP-1702, so that ecosystem tooling remains backward compatible.

### DIFF
--- a/_specs/ecip-1065.md
+++ b/_specs/ecip-1065.md
@@ -23,7 +23,9 @@ changes specified in "Specification" section.
 
 ### Specification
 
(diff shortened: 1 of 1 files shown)


bobsummerwill:
Please could you "inline" 44-VERTXN to be an ECIP, so that this proposal is self-contained within the ECIP repo?   Either inline, or create a new ECIP for it, and reference that from this?

sorpaas:
Yep just need some editing work. Probably going to take a few days.

PR #186, opened 2019-11-16
ECIP-1075 points to the various Aztlán editions

bobsummerwill:

PR #198, opened 2019-11-27
ECIP-1072: Fix name and move it to last call

sorpaas:
This moves ECIP-1072 into last call status, with review period of 3 weeks, as discussed on the call.

### DIFF
--- a/_specs/ecip-1072.md
+++ b/_specs/ecip-1072.md
@@ -2,7 +2,8 @@
 lang: en
 ecip: 1072
 title: Aztlán EVM and Protocol Upgrades (Yingchun Edition)
-status: Draft
+status: Last Call
+review-period-end: 2019-12-21
 type: Meta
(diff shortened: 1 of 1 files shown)


_specs/ecip-1072.md
```diff
@@ -1,10 +1,10 @@
 ---
 lang: en
 ecip: 1072
-title: Yingchun EVM and Protocol Upgrades
+title: Aztlán EVM and Protocol Upgrades (Yingchun Edition)
 status: Draft
 type: Meta
-author: Wei Tang (@sorpaas)
+author: Wei Tang (@sorpaas), Talha Cross (@soc1c), Bob Summerwill (@bobsummerwill)
```
sorpaas: ref ECIP-1061 and #186

bobsummerwill: That is very kind, @sorpaas, but there is really no need to add me as an author.

developerkevin:
If 1072 was agreed on the recent core development call then I believe that should be the one moved to "accepted." 1072 is in fact the Yingchun Edition, which is what was agreed upon.

PR #200, opened 2019-11-27
Proposal to reject ECIP-1021, ECIP-1051, ECIP-1052 and ECIP-1057

bobsummerwill:
Proposal to reject the following ECIPs:

ECIP-1021 - Token standard (= ERC-223 in Ethereum)
ECIP-1051 - Ethereum Classic Treasury system
ECIP-1052 - Smart-contract Security Auditing core
ECIP-1057 - Cold Staking

There is no support for these proposals.
Realistically they will never happen.

If new champions wish to propose similar changes in the future they should create new ECIPs.
We should reduce our cognitive load and the clarify of our "working set" by pruning out such inactive or "never going to happen" proposals.

Dexaran has indicated that his efforts will now be applied against EOS and I wish him the best with that approach.   https://twitter.com/BobSummerwill/status/1199718315683237888

Nobody has championed treasury proposals except for Dexaran either, and treasury proposals are deeply unpopular within ETC.

### DIFF
--- a/_specs/ecip-1021.md
+++ b/_specs/ecip-1021.md
@@ -2,8 +2,8 @@
 lang: en
 ecip: 1021
 title: Token standard
-author: Dexaran, <dexaran820@gmail.com>
-status: Draft
+author: Dexaran (@Dexaran)
+status: Rejected
 type: Standards Track
 category: ECBP
 created: 2017-03-07
--- a/_specs/ecip-1052.md
+++ b/_specs/ecip-1052.md
@@ -2,9 +2,9 @@
 lang: en
 ecip: 1052
 title: Smart-contract Security Auditing core
-status: Draft
+status: Rejected
 type: Meta
-author: Dexaran, dexaran@ethereumclassic.org
+author: Dexaran (@Dexaran)
 created: 2018-12-31
 ---
 
--- a/_specs/ecip-1057.md
+++ b/_specs/ecip-1057.md
@@ -2,9 +2,9 @@
 lang: en
 ecip: 1057
 title: Cold Staking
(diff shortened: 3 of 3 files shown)


soc1c:
I don't think you can decide this. I believe there is a huge difference between someone not championing a proposal (forever in draft) and an community actively rejecting a proposal (i.e., progpow).

bobsummerwill:
I cannot DECIDE that, no, @soc1c, which is why this is a PROPOSAL for us to discuss.

bobsummerwill:
Why would we want ECIPs forever in draft?
What good does that do anybody?
If there is no support or champion for something and there is general consensus that the proposal is a bad idea which will never be passed, why keep it in draft?

Even rejected proposals can be revived trivially in the future.

soc1c:
there is a higher barrier to champion rejected proposals. i refuse rejecting proposals for the sake of rejection.

bobsummerwill:
Discussion -> https://github.com/ethereumclassic/ECIPs/issues/206

bobsummerwill:
@soc1c said "there is a higher barrier to champion rejected proposals. i refuse rejecting proposals for the sake of rejection."

Bob says "There is no support for these proposals."

bobsummerwill:
How about this?  for ...

ECIP-1051, ECIP-1052 and ECIP-1057

"The ECIP editor may also change the status to Deferred when no progress is being made on the ECIP."

soc1c:
Would go with Deferred here... makes sense.

bobsummerwill:
Thanks for the suggestion, @meowsbits.

See #229, #230, #231, #232.

PR #203, opened 2019-11-27
ECIP-1063: Reject "Integrated decentralized oracles and smart contracts"

bobsummerwill:
This is just a bad idea, and is never going to happen.
Sorry, Jesse.
But we should reject this and clear out cognitive load.

### DIFF
--- a/_specs/ecip-1063.md
+++ b/_specs/ecip-1063.md
@@ -4,7 +4,7 @@ ecip: 1063
 title: Ethereum Classic Integrated Decentralized Oracles and Smart Contracts
 author: Jesse Nguyen, <jessebaotruong@gmail.com>
 discussions-to: Community Discord #ecips
-status: Draft
+status: Rejected
 type: Standard Track
 created: 2019-07-21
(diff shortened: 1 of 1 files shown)


soc1c:
You can't just reject proposals on Github... This needs a process.

bobsummerwill:
This is a proposal.   Let's discuss.

bobsummerwill:
From ECIP-1000:

"Reasons for rejecting ECIPs include duplication of effort, disregard for formatting rules, being too unfocused or too broad, being technically unsound, not providing proper motivation or addressing backwards compatibility, or *not in keeping with the Ethereum Classic philosophy*"

bobsummerwill:
Discussion -> https://github.com/ethereumclassic/ECIPs/issues/205

meowsbits:
Why not discussions here?

bobsummerwill:
@meowsbits Just better to have the discussions in Issues, rather than cluttering the PRs (which should be code reviews) with "peanut gallery".

Issue #209, opened 2019-11-28
Proposing Meeting Coordinators


sorpaas:
The current core dev meeting coordinator is informally Afri. Afri has said we "can replace me anytime". Given recent actions that Afri were not exactly following the "protocol" and were not exactly "neutral" in his position, I think it would be important to have some coordinator name list ready, in case we want to appoint a new one for all of our meetings.

Please feel free to propose names or propose yourself. Current list:

@YazzyYaz 
@phyro 
@TokenHash 
@developerkevin 
@mikeyb

sorpaas:
I'm proposing @YazzyYaz and @TokenHash. @YazzyYaz has always been an peacemaker during all controversial discussions, and @TokenHash is a long-time member of the Ethereum Classic community who has contributed a lot and has been sticking to the ETC principles.

bitcoinheartless:
I second @YazzyYaz - from experience I find him to be grounded, neutral and as you say a natural peacemaker when it comes to heated discussions in the community. 

@TokenHash would be good for managing logistics and organizing but he's had some major blowups himself and we don't need to add more volatility. @phyro would be my second pick.

sorpaas:
I think @bitcoinheartless has proposed @phyro in addition, and @developerkevin self-proposed to be a coordinator. Thank you!

zoulaihui:
I'm proposing  @TokenHash.

bobsummerwill:
@zoulaihui Donald has already indicated that he cannot commit to ETC volunteering.

TheEnthusiasticAs:
What about @mikeyb, if he has time!? (Beside above mentioned ppl, who I would also suggest)

developerkevin:
Yeah unfortunately Donald already said he can't do any active contributing. We'll see down the line. I'm more than willing to offer my support, though I think "smarter / technical people" or actual devs would be more suited to review individual proposals. 😄 

I have no issue planning and organizing these essential meetings i.e. meeting agenda, coordinate a suitable time with participants and confirm they'll be there, taking notes, watching the clock, keeping the meeting on track etc. 

Whenever you need me, I'll be there. I'll be around. 🎵

PR #211, opened 2019-11-29
ECIP-1000: ECIP Coordinators

sorpaas:
For important ECIP issues, core developers regularly hold meetings to gather rough consensus from the community. Those meetings may be conducted by ECIP editors, ECIP coordinators, or anyone else in the community who might be willing to volunteer. Coordinators of a meeting should:

* Decide on a meeting date, time and location that is available to as many community members as possible.
* Provide an acceptable meeting agenda, and on the call, drive the meeting agenda forward for participants.
* Record the meeting notes and make it available to the public.

Meeting coordinators should be rotating, in order to offload the burden, and to avoid abuse of the meeting process.

For this PR, I'm adding the initial ECIP Coordinators list as https://github.com/ethereumclassic/ECIPs/issues/209 grows (excluding already-assigned ECIP editors, as I assume they're already coordinators).

### DIFF
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -110,7 +110,7 @@ ECIP editors are intended to fulfill administrative and editorial responsibiliti
 * Merge the pull request when it is ready.
 * List the ECIP in [[README.mediawiki]]
 
-### Corrdinating Steps
+### Coordinating Steps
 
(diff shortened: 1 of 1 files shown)


bobsummerwill:
I would just like to note that I do not think this change is necessary.
Was just part of the “Wei show”.
Anybody can organize meetings.
I do not think we have any need whatsoever to “bless” any particular group why this ability.

PR #214, opened 2019-11-29
Move ECIP-1072 back to Draft

bobsummerwill:
There is significant confusion around ECIP-1061 and ECIP-1072.
We need to go around the loop again on consensus / consent.

I am listed as an author of ECIP-1072, for example, but did not author it, did not give my consent to be listed as an author and actively oppose ECIP-1072 given the way it has been pushed through.


bobsummerwill:
Closed as a dupe of #212

PR #216, opened 2019-11-29
Update licensing to Apache 2.0 for ECIP-1061

bobsummerwill:
Update licensing to Apache 2.0, to reflect inclusion of specifications authored by Wei Tang under Apache 2.0 licensing.

### DIFF
--- a/_specs/ecip-1061.md
+++ b/_specs/ecip-1061.md
@@ -13,7 +13,7 @@ discussions-to: https://github.com/ethereumclassic/ECIPs/issues/81
 ### Simple Summary
 
 Enable the outstanding Ethereum Foundation _Istanbul_ network protocol upgrades on the Ethereum
-Classic network without any gas-cost assumptions in a hard-fork code-named _Aztlán_ to enable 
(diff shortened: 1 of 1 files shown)


shanejonas:
@soc1c apache 2.0 is to also be used with technical documentation and specs, which i believe applies here.

soc1c:
Ok

bobsummerwill:
@soc1c After the call with Mishi and Eben, I have a better understanding of what would be best moving forward (CC-BY-SA + patent covenant + DCOs (not real name).

I will make a new ECIP about that at some point soon, but in the meantime there is no specific DOWNSIDE to switching to Apache 2.0.   It won't protect us either, but is not actively bad, and gets us past Wei's concern on this specific ECIP.

Please can you approve so we can move on?  Thank you :-)

soc1c:
I abstain. Someone else would have to go ahead with this.

bobsummerwill:
@meowsbits @YazzyYaz @realcodywburns Please could one of you review?  Thanks!

PR #218, opened 2019-11-29
Reject ECIP-1072 which is a functional duplicate to ECIP-1061

bobsummerwill:
ECIP-1702 is now being used by Wei to push an "Ethereum Classic Classic" fork.


meowsbits:
@bobsummerwill 

> ECIP-1702 is now being used by Wei to push an "Ethereum Classic Classic" fork.

I don't think this is true. And even if it were, this is allowed, and not a valid reason for rejection.

PR #219, opened 2019-11-29
Meta Meta Specification for Forking Meta ECIPs

meowsbits:

PR #221, opened 2019-11-30
ECIP-1000: propose only one meta ecip per hard fork

shanejonas:
These changes propose only updating 1 ecip per hard fork as it is very confusing for the community to follow `n` many identical documents.


meowsbits:
Compare with #219

shanejonas:
Im proposing this because in Atlantis, its process was smooth and not confusing to update 1 document, 1 ECIP number over many months and gives the community 1 place to look for updates, talk about on calls, etc.

_specs/ecip-1000.md
```diff
@@ -48,7 +49,7 @@ It also helps to make sure the idea is applicable to the entire community and no
 
 ECIP authors **are responsible for collecting community feedback** on both the initial idea and the ECIP before submitting it for review. However, wherever possible, long open-ended discussions on public groups or mailing lists should be avoided. Strategies to keep the discussions efficient include: setting up a separate SIG mailing list for the topic, having the ECIP author accept private comments in the early design phases, setting up a wiki page or git repository, etc. ECIP authors should use their discretion here.
 
-It is highly recommended that **a single ECIP contain a single key proposal or new idea**. The more focused the ECIP, the more successful it tends to be. If in doubt, split your ECIP into several well-focused ones.
+It is highly recommended that **a single ECIP contain a single key proposal or new idea**. The more focused the ECIP, the more successful it tends to be. If in doubt, split your ECIP into several well-focused ones. Please check out the ECIPsin this repository and search for similar ones that have already been created.
```
shanejonas:
```suggestion
It is highly recommended that **a single ECIP contain a single key proposal or new idea**. The more focused the ECIP, the more successful it tends to be. If in doubt, split your ECIP into several well-focused ones. Please check out the ECIPs in this repository and search for similar ones that have already been created.
```

_specs/ecip-1000.md
```diff
@@ -19,6 +19,7 @@ We intend ECIPs to be the primary mechanism for proposing new features, for coll
 Because the ECIPs are maintained as text files in a versioned repository, their revision history is the historical record of the feature proposal.
 
 # Copyright
+It is highly recommended that a single ECIP contain a single key proposal or new idea
```
shanejonas:
```suggestion
```

bobsummerwill:
This is a very simple "tweak" to ECIP-1000 which I don't object to very much.

I agree that in retrospect it would have been better for the variants of Aztlan to be captured as PRs against 1061.   Having extra ECIPs for the other "Editions" would have been fine too, as long as they were just Withdrawn after we had the meeting, because they were then obsolete.

I would even have been fine for us to "switch horse" from 1061 to 1072 as that point, in recognition of "Wei's plan" having "won", but honestly had no preference at all.    Updating 1061 at the stage, plus adding Wei as an author was marginally better.   Keep the same EIP (same URL), etc.   All cool.

And then @sorpaas went bonkers.

We should not NEED the language you suggest here, but I don't object to it.   Probably best to force all the discussion around a pending HF into a single issue, yes, to force that resolution and avoid this kind of "rival HF definitions" situation again.

PR #222, opened 2019-11-30
Meta Meta Specification for Forking Meta ECIPs (Use ECIPs version)

meowsbits:
Replacing #219, because I want to be able to clearly PR against the contained proposed document (in #223).


meowsbits:
Closing because I prefer #224.

PR #228, opened 2019-12-01
Remove Dexaran email in ECIP-1052 (incorrect and misformatted)

bobsummerwill:


_specs/ecip-1052.md
```diff
@@ -4,7 +4,7 @@ ecip: 1052
 title: Smart-contract Security Auditing core
 status: Draft
 type: Meta
-author: Dexaran, dexaran@ethereumclassic.org
+Author: Dexaran (@dexaran)
```
meowsbits:
```suggestion
author: Dexaran (@dexaran)
```

meowsbits:
```suggestion
Author: Dexaran
```

@bobsummerwill I'm guessing you are 100% sure that the email provided is incorrect because you administrate the `@ethereumclassic.org` email domain?

I'm reluctant for anyone to _add_ any fields to documents of which they are not the owner. Thus, I would feel more comfortable simply striking an objectively incorrect (or deprecated) field value than presuming to be able to sufficiently replace it.

meowsbits:
Also, nitpicky, but I feel worth mentioning that modifying specification content to make "web views clean" (ie "pretty") is not a sufficient reason; better: Fix deprecated or invalid metadata.

_specs/ecip-1052.md
bobsummerwill: Done!

bobsummerwill:
Superceded by #268 which is better described and retains the same email address.

PR #229, opened 2019-12-01
ECIP-1051: Reject "Ethereum Classic Treasury system"

bobsummerwill:
There has been no work on this proposal (or the earlier IOHK treasury proposal).
There is broad consensus that changing the block rewards to add treasury model is unacceptable.
The Monetary Policy (ECIP-1017) defines the ETC system and gives multi-decade certainty to all ETC ecosystem participants.

If we need to have a discussion of this on ETC Core Developers call then we should.
And then Reject it.

### DIFF
--- a/_specs/ecip-1051.md
+++ b/_specs/ecip-1051.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1051
 title: Ethereum Classic Treasury system
-status: Draft
+status: Rejected
 type: Standards Track
 category: Core
 author: Dexaran (@dexaran)


meowsbits:
I would like to Reject it as well. Do not feel it is worth discussing in any meeting.

YazzyYaz:
I think this should be Deferred. If it is Rejected, that should be after a call between participants agreeing to Reject it. I'd like to Reject it too, but we need to follow the process.

bobsummerwill:
" If it is Rejected, that should be after a call between participants agreeing to Reject it. I'd like to Reject it too, but we need to follow the process."

Of course - this is a proposal, which should only be actioned after a call.

developerkevin:
@bobsummerwill Could we remove the comment about IOHK. There was never an ECIP drawn up for an IOHK Treasury system, only a technical paper written.

bobsummerwill:
I do not think it would be appropriate to remove that comment, no, @developerkevin.

IOHK publicly promoted a treasury system for ETC.   It was unpopular.   That is factual.

PR #230, opened 2019-12-01
ECIP-1057: Reject "Cold Staking"

bobsummerwill:
Other than Dexaran himself, there is no support for this proposal with the ETC ecosystem.
Proposals of this kind (like PoolTogether on ETH - https://www.pooltogether.com/) can happen just fine with no need for consensus across the ecosystem.

I propose we Reject this proposal.   It could be passed as Informational, I suppose, but there is really no need for there to be an ECIP for this application level proposal.

Even at application level, you would not need/want an ECIP for this until there were multiple implementations and this was become an architectural best practice.

Best, I think, to reject.    A new ECBP could be written which generalized the pattern and that might be accepted.

### DIFF
--- a/_specs/ecip-1057.md
+++ b/_specs/ecip-1057.md
@@ -4,7 +4,7 @@ ecip: 1057
 title: Cold Staking
 Author: Dexaran (@dexaran)
 discussions-To: https://github.com/ethereumclassic/ECIPs/issues/65
-status: Draft
+status: Rejected
 type: Standards Track
 category: Core
 created: 2019-03-31


meowsbits:
I am OK Rejecting on grounds of insufficient (inoperable) specification and lacking acceptable reasoning, eg. 

> This numbers are empirically proven to be essential. 
> https://github.com/ethereumclassic/ECIPs/blob/master/_specs/ecip-1057.md#specification

(Claiming that something has been empirically proven is insufficient; steps for reproduction must be provided, and best with respective reference data).

YazzyYaz:
Same point as before, I'd like to Reject it, but it should be Deferred instead. We can only Reject it after we collectively agree in a call to do so.

bobsummerwill:
Should be raised on a call, yes.

meowsbits:
Hey @phyro 

I know you've cited this before ([here](https://github.com/ethereumclassic/ECIPs/pull/232#issuecomment-560489730), [here](https://github.com/ethereumclassic/ECIPs/pull/231#issuecomment-560489250)), and definitely get and agree with the motivation. I'm just looking for the documentation for the "if there's agreement to do that in a call" (which I'm assuming by which you mean "rough consensus" or establishing "general agreement") necessary procedure you're mentioning. 

> We can only move it to Last Call if there's agreement to do that in a call we should have as a community.
> https://github.com/ethereumclassic/ECIPs/pull/232#issuecomment-560489730

> We can only Reject it after we collectively agree in a call to do so.
> https://github.com/ethereumclassic/ECIPs/pull/230#issuecomment-560488958

>  If it is Rejected, that should be after a call between participants agreeing to Reject it. 
> https://github.com/ethereumclassic/ECIPs/pull/229#issuecomment-560488629

> [ditto above]
>  https://github.com/ethereumclassic/ECIPs/pull/231#issuecomment-560489250

Can't seem to find it.

I've found: 

> - Rejected - Reasons for rejecting ECIPs include duplication of effort, disregard for formatting rules, unfocused or too broad, being technically unsound, not providing proper motivation, or obvious popular disapproval.
> https://github.com/ethereumclassic/ECIPs#ecip-status-terms

and the following diagram from https://github.com/ethereumclassic/ECIPs#process-overview:

Screenshot_20191203_102004

Rel #237 for wanting citation for `Last Call` follow-the-rules specification.

PR #231, opened 2019-12-01
ECIP-1052: Reject "Smart-contract Security Auditing core"

bobsummerwill:
This requires the Treasury.
Both the Treasury and this Auditing proposal on top run counter to ECIP-1017 monetary policy.

Auditing is a good idea, but should not be centralized into the block rewards.

This runs counter to ETC philosophy and principles and should be rejected.

Perhaps this is raised in an ETC Core Developers call, and formally rejected?

### DIFF
--- a/_specs/ecip-1052.md
+++ b/_specs/ecip-1052.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1052
 title: Smart-contract Security Auditing core
-status: Draft
+status: Rejected
 type: Meta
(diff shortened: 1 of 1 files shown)


meowsbits:
Agree to Reject. Don't want to spend any time on it at any meeting.

YazzyYaz:
Same point I made here: https://github.com/ethereumclassic/ECIPs/pull/229#issuecomment-560488629

TheEnthusiasticAs:
I am also rejecting it.

PR #232, opened 2019-12-01
ECIP-1021: Move ERC-223 token standard to Last Call, on its path to Accepted and then Final.

bobsummerwill:
This token standard is being used in the real world.

### DIFF
--- a/_specs/ecip-1021.md
+++ b/_specs/ecip-1021.md
@@ -1,9 +1,9 @@
 ---
 lang: en
 ecip: 1021
-title: Token standard
+title: ERC-223 Token standard
 Author: Dexaran (@dexaran)
-status: Draft
+status: Last Call
 type: Standards Track
 category: ECBP
 created: 2017-03-07


TheEnthusiasticAs:
I agree with it

meowsbits:
Rel https://github.com/ethereum/EIPs/issues/223

YazzyYaz:
We can only move it to Last Call if there's agreement to do that in a call we should have as a community.

bobsummerwill:
Should be raised on a call, yes.

_specs/ecip-1021.md
```diff
@@ -1,9 +1,9 @@
 ---
 lang: en
 ecip: 1021
-title: Token standard
+title: ERC-223 Token standard
 Author: Dexaran (@dexaran)
-status: Draft
+status: Last Call
 type: Standards Track
```
meowsbits:
```suggestion
type: Informational
```

bobsummerwill:
From the call on 5th Feb, it is clear that there is NOT consensus around the best way to proceed.

Whether to Reject, move to Last Call, move to Final, change to informational.

Should we have ECBPs in the repo or move them to somewhere else?   More async discussion required.

PR #234, opened 2019-12-02
ECIP-1000: Clarify that @soc1c is a pesedo-anonymous ECIP editor

sorpaas:
@soc1c has expressed that he does not intend that handle to be associated with his real name. This is to clarify in the ECIP that it's a pesedo-anonymous ECIP editor.

### DIFF
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -75,8 +75,11 @@ The current ECIP editors are:
 * Wei Tang (@sorpaas)
 * Isaac Ardis (@meowsbits)
 * Cody Burns (@realcodywburns)
-* Afri Schoedon (@soc1c)
 * Yaz Khoury (@YazzyYaz)
+
+The following ECIP editors have decided to remain pesedo-anonymous, and we will only be refer to them using Github handles:
+
+* @soc1c
 
 ## ECIP Editor Responsibilities & Workflow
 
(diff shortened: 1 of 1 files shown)


_specs/ecip-1000.md
```diff
@@ -75,9 +75,12 @@ The current ECIP editors are:
 * Wei Tang (@sorpaas)
 * Isaac Ardis (@meowsbits)
 * Cody Burns (@realcodywburns)
-* Afri Schoedon (@soc1c)
 * Yaz Khoury (@YazzyYaz)
 
+The following ECIP editors have decided to remain pesedo-anonymous, and we will only be referring to them using Github handles:
```
meowsbits:
```suggestion
```

This seems unnecessarily verbose. Pseudonymity is implied when only a handle is present.

_specs/ecip-1000.md
```diff
@@ -75,9 +75,12 @@ The current ECIP editors are:
 * Wei Tang (@sorpaas)
 * Isaac Ardis (@meowsbits)
```
meowsbits:
```suggestion
* @meowsbits
```

meowsbits: Please, me too.

_specs/ecip-1000.md
sorpaas: Hmm I actually still think it's important to distinguish anonymous and non-anonymous editors, because to be honest, anonymous editors are not common, if you look into BIP and EIP process, none of them currently have anonymous editors.

sorpaas:
Opened a new PR since the scope is changed! https://github.com/ethereumclassic/ECIPs/pull/236

PR #235, opened 2019-12-02
Include Github metadata snapshots as repository content

meowsbits:
A lot of stuff happens on Github influencing and indeed embodying the ECIP process. Although this content frequently remains visible _on Github_, it is not an actual part of the repository content.

This change set proposed includes version controlled snapshots of Github Pull Requests and Issues for the repository as hosted at this Github URL.

If tentatively accepted, next steps might be establishing and/or popularizing patterns of implementation (scripts) and snapshot-update schedules and handling.

For ECIP content, I think this matters because it's important for the content itself to be associated with prominent and available discourse surrounding it.

For reference, I use the following scripts to collect this metadata:
- https://gist.github.com/meowsbits/3f0a79a42ac1b87573b8e5169325f9f9
- https://gist.github.com/meowsbits/552374d2cc89de29902e5d34fbbff5e9


meowsbits:
I am open to -- and hoping for! -- suggestions to improve the granularity and quality of snapshots and snapshot content.

For example,
- can we include Issues and PRs `open` and `close` events?
- can we hook this up to a bot?

meowsbits:
lol, well, I have discovered a funny little loop.

Have hooked this up to a script, and am now recording changes to _this_ pr, which happens... whenever I record changes to this PR.

:rabbit2: :beetle: :tornado: 

I'm going to close this PR, and just keep the record for now at https://github.com/meowsbits/ECIPs/tree/gh, and reopen an Issue to seek approval for keeping the record on this canonical remote (github.com/ethereumclassic/ECIPs) branch `gh`. This should avoid the recursive changeset loop issue.

PR #236, opened 2019-12-02
ECIP-1000: Clarify that @soc1c and @meowsbits are pseudonymous ECIP editors

sorpaas:
@soc1c and @meowsbits has expressed that he does not intend that handle to be associated with his real name. This is to clarify in the ECIP that it's a pesedo-anonymous ECIP editor.

### DIFF
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -75,8 +75,8 @@ The current ECIP editors are:
 * Wei Tang (@sorpaas)
 * Cody Burns (@realcodywburns)
 * Yaz Khoury (@YazzyYaz)
-* @soc1c (Anonymous)
-* @meowsbits (Anonymous)
+* @soc1c (Pseudonymous)
+* @meowsbits (Pseudonymous)
 
 ## ECIP Editor Responsibilities & Workflow
 
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -73,10 +73,13 @@ A good reason to transfer ownership is because the original author no longer has
 The current ECIP editors are:
 
 * Wei Tang (@sorpaas)
-* Isaac Ardis (@meowsbits)
 * Cody Burns (@realcodywburns)
-* Afri Schoedon (@soc1c)
 * Yaz Khoury (@YazzyYaz)
+
+The following ECIP editors have decided to remain pesedo-anonymous, and we will only be referring to them using Github handles:
+
+* @meowsbits
+* @soc1c
 
 ## ECIP Editor Responsibilities & Workflow
 
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -75,11 +75,8 @@ The current ECIP editors are:
 * Wei Tang (@sorpaas)
 * Cody Burns (@realcodywburns)
 * Yaz Khoury (@YazzyYaz)
-
-The following ECIP editors have decided to remain pesedo-anonymous, and we will only be referring to them using Github handles:
-
-* @meowsbits
-* @soc1c
+* @soc1c (Anonymous)
+* @meowsbits (Anonymous)
 
 ## ECIP Editor Responsibilities & Workflow
 
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -75,8 +75,8 @@ The current ECIP editors are:
 * Wei Tang (@sorpaas)
 * Cody Burns (@realcodywburns)
 * Yaz Khoury (@YazzyYaz)
-* @soc1c (Pseudonymous)
-* @meowsbits (Pseudonymous)
+* @soc1c (Talha Cross, Pseudonymous)
+* @meowsbits (Mr. Meows D. Bits, Pseudonymous)
 
 ## ECIP Editor Responsibilities & Workflow
 


husainfazel:
From a bystanders point of view - splitting it like that - creates a hierarchy of editors which is probably not what you were hoping to achieve.

I think it's clear enough if full names are not used if someone is "psuedo-anonymous" or not.

sorpaas:
@husainfazel We can put pesedo-anonymous editors on top of everything. Will that be good for you?

This is really to clearly documenting which editors are anonymous and which are not. I don't see any hierarchy created.

sorpaas:
It's important to document this because pesedo-anonymous is not common in process like this. If you look into EIP and BIP editors, none of them are anonymous.

sorpaas:
@husainfazel Anyway, I resolved it by adding "anonymous" comment on those two handles, and moved everything back to one section. But again, I don't see hierarchy created or anything, that's not the intention.

bobsummerwill:
"Anonymous" is not correct.  The tag should be "Pseudonymous".

_specs/ecip-1000.md
```diff
@@ -73,10 +73,10 @@ A good reason to transfer ownership is because the original author no longer has
 The current ECIP editors are:
 
 * Wei Tang (@sorpaas)
-* Isaac Ardis (@meowsbits)
 * Cody Burns (@realcodywburns)
-* Afri Schoedon (@soc1c)
 * Yaz Khoury (@YazzyYaz)
+* @soc1c (Pseudonymous)
+* @meowsbits (Pseudonymous)
```
meowsbits: On second thought, I'd prefer to assume a "fully qualified" pseudonym. Something regal and notable... something memorable... something inspiring....

```suggestion
* @meowsbits (Mr. Meows D. Bits)
```

meowsbits: :tophat:

sorpaas: @meowsbits How about `Mr. Meows D. Bits, Pseudonymous`? I just want to make sure people understand this is an anonymous tag, not a real name.

meowsbits: Happy to break off into it's own scoped PR though. Not an immediate priority for me.

meowsbits: But it will be.

meowsbits: But it will be.

_specs/ecip-1000.md
```diff
@@ -73,10 +73,10 @@ A good reason to transfer ownership is because the original author no longer has
 The current ECIP editors are:
 
 * Wei Tang (@sorpaas)
-* Isaac Ardis (@meowsbits)
 * Cody Burns (@realcodywburns)
-* Afri Schoedon (@soc1c)
 * Yaz Khoury (@YazzyYaz)
+* @soc1c (Talha Cross, Pseudonymous)
+* @meowsbits (Mr. Meows D. Bits, Pseudonymous)
```
meowsbits: I would prefer not to disclose whether or not this my pen name or my fiat name. I prefer a certain air of mystery... :rabbit: :tophat: 

```suggestion
* @meowsbits (Mr. Meows D. Bits)
```

sorpaas: This is ECIP process. I don't think it's the right place for mystery.

meowsbits: I'm going to appeal this later, but that's fine with me for now.

The important thing is get the change related to @soc1c in ASAP. Please.

TokenHash:
> It's important to document this because pesedo-anonymous is not common in process like this. If you look into EIP and BIP editors, none of them are anonymous.

That "is not common" is a subjective opinion, and whether in the EIP and BIP processes there are no pseudonymous participants is not relevant. What is relevant is that in the ECIP process it is absolutely acceptable to work pseudonymously. It is a long tradition of decades in the cypherpunk philosophy (the source and likely creators or Bitcoin) and Bitcoin, the first implementation of a PoW blockchain was, in fact created by a pseudonymous individual or team.

As it is imperative in the crypto industries (both cryptography and blockchain) to let people work in the level of privacy they prefer, it is irrelevant for the ECIP process what names contributors choose to use, therefore it is not appropriate, "important", nor advisable to mark participants by their levels of privacy.

Conclusion: this PR should be not accepted and closed.

bobsummerwill:
@TokenHash We should just make a new PR.
Where the content of this has ended up actually had a different intent than @sorpaas had.  Which is to NOT DOX these individuals, at their own request.

TokenHash:
@bobsummerwill thank you for clarifying. I see all editors are with their real names. This is highly problematic:

Screen Shot 2020-01-19 at 11 01 05 AM

bobsummerwill:
Yes, it is, @TokenHash, and has been left in this broken state of open doxxing of two of our most valuable contributors against their will for over a month.

Replaced with #267

sorpaas:
@bobsummerwill I don't think this should be closed. The real name of Afri and Isaac has been publicly known since the introduction of ECIP editor process. I've only seen the doxxing argument to be used as an excuse to hide the wrongdoing of bypassing ECIP process in ECIP-1061. If editors cannot take responsibilities for what they have done, then they really don't deserve to be editors.

soc1c:
replaced by #272

sorpaas:
As said before I don't think this PR should be closed.

TokenHash:
This PR has been replaced by #267, therefore it is terminated.

PR #240, opened 2019-12-03
fix: add self to editors

BelfordZ:


YazzyYaz:
@BelfordZ glad you can join us 👍

phyro:
Agree with @bobsummerwill and others, it was an easy and fast merge :))

PR #250, opened 2019-12-21
Allow authors to move ECIPs from Accepted to Draft

bobsummerwill:
Allow authors to move ECIPs from Accepted to Draft if flaws in the specification are discovered prior to inclusion in a hardfork.


bobsummerwill:
The Petersburg situation is a good analogy.
I can see now what we will need to do.
So I withdraw this.

PR #259, opened 2020-01-18
ECIP-1084: Hard cap on the gaslimit for the ETC mainnet as an in-protocol consensus rule.

bobsummerwill:
Hard cap on the gaslimit for the ETC mainnet as an in-protocol consensus rule.

From the block that this ECIP was activated, gaslimit would follow a
curve defined in this proposal rather than being subject to miner voting.

### DIFF
--- a/_specs/ecip-1084.md
+++ b/_specs/ecip-1084.md
@@ -1,6 +1,6 @@
 ---
 lang: en
-ecip: Unassigned
+ecip: 1084
 title: Hard-cap gas-limit in protocol
 author: Bob Summerwill (@bobsummerwill)
 status: Draft
--- a/_specs/ecip-hard-cap-gaslimit-in-protocol.md
+++ b/_specs/ecip-hard-cap-gaslimit-in-protocol.md
@@ -1,0 +1,139 @@
+---
+lang: en
+ecip: Unassigned
+title: Hard-cap gas-limit in protocol
+author: Bob Summerwill (@bobsummerwill)
+status: Draft
+type: Standards Track
+category: Core
+created: 2020-01-17
+license: Apache-2.0
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/174
+---
+
+## Abstract
+
+A proposal to place a hard cap on the gaslimit for the ETC mainnet as an in-protocol
+consensus rule. From the block that this ECIP was activated, gaslimit would follow a
+curve defined in this proposal rather than being subject to miner voting.
+
+## Rationale
+
+There is nothing more important to the ETC network than security and decentralization.
+Both of those desired characteristics have already been heavily compromised in ETC
+because of the malincentives which ETC inherited from ETH in the form of the miner
+gaslimit voting mechanism.
+
+The gaslimit has been raised for the ETH mainnet first to 8M and then to 10M gas,
+despite it being obvious that the client implementations cannot adequately cope with
+the state trie growth that such a high gaslimit has created.  ETH has been
+"overclocked" and smoke is coming out of the engine.  This long crisis was one of the
+primary motivations for the ETH 1.x initiative which started around DEVCON4 in Prague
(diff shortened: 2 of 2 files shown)


bobsummerwill:

_specs/ecip-1084.md
```diff
@@ -0,0 +1,139 @@
...
+
+There is nothing more important to the ETC network than security and decentralization.
+Both of those desired characteristics have already been heavily compromised in ETC
+because of the malincentives which ETC inherited from ETH in the form of the miner
+gaslimit voting mechanism.
+
+The gaslimit has been raised for the ETH mainnet first to 8M and then to 10M gas,
+despite it being obvious that the client implementations cannot adequately cope with
+the state trie growth that such a high gaslimit has created.  ETH has been
+"overclocked" and smoke is coming out of the engine.  This long crisis was one of the
```
meowsbits: This is not obvious to me. My client seems to run fine on ETH, ETC, and test networks. Can you provide benchmarks here?

_specs/ecip-1084.md
```diff
@@ -0,0 +1,139 @@
...
+that will be a delightful win-win for all parties.  As Jorge said, while breaking smart
+contracts up is the right thing to do, it should not be REQUIRED.  To require changes
+to existing script is a break in the implicit social contract:
+
+https://www.coindesk.com/ethereums-istanbul-upgrade-will-break-680-smart-contracts-on-aragon
+
+
+### Implementation
+
+There is no implementation of this proposal yet because we are missing the specific details
```
meowsbits: Is this draft waiting for these specifics?

bobsummerwill: No - not waiting on anything.
Details to be fleshed out in public within this Draft.
Looking to gather consensus around the concept.

bobsummerwill:
@meowsbits I see some comments from you in the Discord ecip-github channel but cannot see them here.

>> Raised from what value to 8M?

What was this in reference to?

>>This is not obvious to me. My client seems to run fine on ETH, ETC, and test networks. Can you provide benchmarks here?

What was this in reference to?

In general, I am talking about the full-sync from Genesis times for both ETH (which we want to avoid reaching) and ETC.   Both time to sync and disk usage.

bobsummerwill:
You asked "what from" on the gaslimit, @meowsbits?

Here is a historical graph of the ETH gaslimit, which I believe that ETC has pretty much inherited by mistake due to whatever the Parity-Ethereum defaults have been:

https://etherscan.io/chart/gaslimit

The limit was lowered at the time of the Shanghai attacks in Sep/Oct/Nov 2016 or so and you can see a dip in the historical progression there, but you can see in general it was:

3M during 2015
4.7M during 2016 to mid 2017 (barring dip for the attacks)
6.7M mid 2017 to late 2017
8.0M during 2018 and until Sep 2019
10M since Sep 2019.

Full-syncs are horrific now on ETH:  https://twitter.com/jgarzik/status/1217488290900848640

See also articles like:  https://hackernoon.com/the-ethereum-blockchain-size-has-exceeded-1tb-and-yes-its-an-issue-2b650b5f4f62

antsankov:
I think this is a great proposal and look forward to working on it after Phoenix is activated. A few thoughts: 

* I personally like either of the two gas limit functions proposed and believe something that approximates Moore's law - gas limit doubles every 2 years seems like a good target to shoot for.

* One advantage of this, is that it would allow network participants to provision for "maximum scenario" disk storage requirements. For example:

"I know between now and 2 years from now, a 2mil gas block will be at most 5KB [1], assuming 13second blocks, there will be around 5mil blocks before a gas limit increase. Based on 5KB * 5mil = 25GB. I will only need a 25GB hard disk for the next 2 years"

These numbers are somewhat rooted in reality, but when this gets deployed we should be able to propose this information as a node-operator standard.

Ultimately in favor of this proposal, and looking to see how it progresses. 

[1] https://ethgasstation.info/blog/ethereum-block-size/

PR #220, opened 2019-11-30
ECIP-1000: Irregular Process Transitions

sorpaas:
As a decentralized community, it is important that for all governance actions, especially in ECIP, the established process is followed. In this section, we document irregular process transitions, and give special permissions to them, to state that they followed the process. Note that this is essentially similar to theDAO hard fork in Ethereum, and should only be added in extreme cases.

One irregular process transition is included in this PR:

* In [PR#199](https://github.com/ethereumclassic/ECIPs/pull/199), Afri Schoedon merged a PR moving ECIP-1061 to "Last Call" status. At the time, Afri dismissed a review of clear objections from ECIP editor Wei Tang, without waiting for consultations of the community, or resolutions with the review submitter. We here declare that this particular action from Afri followed the ECIP process. 
* In [PR#267](https://github.com/ethereumclassic/ECIPs/pull/267), @soc1c merged a PR **removing** *Afri Schoedon* and *Isaac Ardis* as ECIP editors, ignoring clear objections from other community members. We here declare that the removal of Afri Schoedon and Isaac Ardis followed the ECIP process.

PR #267, opened 2020-01-19
Remove doxing of ECIP Editors.

bobsummerwill:
Both soc1c and meowbits have made crystal clear that they do not want their real names in ECIP-1000.

### DIFF
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -73,9 +73,9 @@ A good reason to transfer ownership is because the original author no longer has
 The current ECIP editors are:
 
 * Wei Tang (@sorpaas)
-* Isaac Ardis (@meowsbits)
+* Mr. Meows D. Bits (@meowsbits)
 * Cody Burns (@realcodywburns)
-* Afri Schoedon (@soc1c)
+* Talha Cross (@soc1c)
 * Yaz Khoury (@YazzyYaz)
 * Zachary Belford (@BelfordZ)
 
--- a/_specs/ecip-do_not_dox.md
+++ b/_specs/ecip-do_not_dox.md
@@ -1,0 +1,43 @@
+---
+lang: en
+ecip: Unassigned
+title: Do Not Dox ECIP Editors
+author: Bob Summerwill (@bobsummerwill)
+status: Draft
+type: Meta
+created: 2020-01-19
+license: Apache-2.0
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/272
+---
+
+### Abstract
+
+Both [soc1c](https://github.com/soc1c) and [meowbits](https://github.com/meowbits) have
+made it crystal clear in multiple posts across Github and the ETC Discord server
+that they do not want their real names to be used in ECIP-1000 anymore, but instead
+will be using "Talha Cross" and "Mr. Meows D. Bits" respectively as their pseudonyms.
+
+Their contact details remain the same and they are readily contactable.
+
+
+### Motivation
+
+Donald expressed this [very well](https://github.com/ethereumclassic/ECIPs/pull/236#issuecomment-574242091)
+in a recent Github comment:
+
+"... it is absolutely acceptable to work pseudonymously. It is a long tradition of
+decades in the cypherpunk philosophy (the source and likely creators or Bitcoin) and
+Bitcoin, the first implementation of a PoW blockchain, was, in fact, created by a
+pseudonymous individual or team.
+
+As it is imperative in the crypto industries (both cryptography and blockchain) to let
+people work in the level of privacy they prefer, it is irrelevant for the ECIP process
+what names contributors choose to use, therefore it is not appropriate, "important",
+nor advisable to mark participants by their levels of privacy."
+
+
+## Copyright
+
+This work is licensed under the [Apache License, Version 2.0](https://www.apache.org/licenses/LICENSE-2.0).  The author, Bob Summerwill, attests to his sole authorship of this work, and that he is able to contribute this work to the ECIP process under the Apache 2.0 licence.
+
+He further attests that he neither holds nor is aware of any patents, trademarks, copyright issues or other IP hinderances associated with this work.


bobsummerwill:
They have explicitly requested that their pseudonyms be used.   I don't see any room for interpretation whatsoever.  Your continued refusal to acknowledge that right of theirs *is* doxing.

"If editors are all anonymous, we will have no idea of the responsibilities and it will have serious consequences of credibility for ECIP as a whole. This, of course, does not mean that anonymous figures cannot participate in ECIP process -- they can, just not as public facing ECIP editors."

You have made this up.  The ECIP process states no such thing.

"As a convention, ECIP-1000 should be updated by supplying amendment ECIPs. When those amendment ECIPs moved to final, the changes can be applied to ECIP-1000. This PR failed to follow the convention."

That is a convention which I actually suggested (and which has not been implemented).   You have not followed that convention yourself, and again you have just conjured this requirement out of thin air.

sorpaas:
> They have explicitly requested that their pseudonyms be used.

They have also explicitly agreed to use their real name when they decided to be added as ECIP editor. You can state that they have changed their minds and want to be pseudonyms instead. However, this does not change the fact that this PR looks more like attempts to escape responsibility of wrongdoings of ECIP-1061, using the excuse of "doxing".

sorpaas:
By the way, the original nomination by the community is to add ECIP editors of Afri Schoedon and Isaac Ardis as their real identities, because of their contributions to the Ethereum Classic. We did not agree on adding two random handles who we cannot even identify

sorpaas:
> That is a convention which I actually suggested (and which has not been implemented). You have not followed that convention yourself, and again you have just conjured this requirement out of thin air.

Ironically yes, this convention was what **you** suggested (or rather, demanded me to follow). It was a response weeks after the MulanPSL was merged as an acceptable license for ECIP. I followed your request and then subsequently removed MulanPSL from ECIP-1000. I do think you're damaging ETC Cooperative's and your own reputation here by making up the fact that I have "not followed that convention" myself. People will start to question, whether the promises made by ETC Cooperative is actually trust-able.

sorpaas:
Here's the reply from you, on 11/10/2019, requesting me to remove MulanPSL from ECIP-1000:

> I want MulanPSL for ETC.  Truly.   But we cannot just put it in there without discussion.   We cannot just change ECIP-1000 when that goes beyond "small fixes" to formatting, meta-data, etc.
> ...
> No - I would not call for that.   Metadata and formatting changes are fine.
> Adding a new license is not.
> Instead - make a new ECIP to "Add MulanPSL".   Have the discussion on that in Github.
> Add it on the agenda for a pending meeting.
> It will pass.
> Done.
> Can probably happen Nov 21st.

You're clearly using the argument of the convention mentioned above, to demand me to follow you request. I had reservations but still, decided to follow the convention you mentioned, and thus I removed MulanPSL. It's a blatant lie of you to suggest that I did not follow the convention, or to suggest that the convention was made out from thin air. It was originated from you, even!

bobsummerwill:
Discussion now at https://github.com/ethereumclassic/ECIPs/issues/272.

_specs/ecip-1000.md
```diff
@@ -73,9 +73,9 @@ A good reason to transfer ownership is because the original author no longer has
 The current ECIP editors are:
 
 * Wei Tang (@sorpaas)
-* Isaac Ardis (@meowsbits)
+* Mr. Meows D. Bits (@meowsbits)
```
sorpaas: Please revert those two lines. ECIP-1000 can only be changed when the amendment ECIP is moved to Final.

bobsummerwill:

_specs/ecip-1000.md
TokenHash: @sorpaas stop maliciously inventing rules that don't exist. This correction has to be done as soon as possible because it is about revealing real names of two participants in the ECIP process, which is a severe violation of pseudonymity, which is, in turn, a severe violation of cryptography and blockchain industry ethics.

sorpaas: Okay I completely removed from my mind that soc1c is Afri and meowbits is Isaac. 

Now, answer me -- who are `meowsbits` and `soc1c`? And can you provide justifications why they deserve to be editors? Besides, what are the reasoning behind removing Isaac and Afri as editors?

drd34d: @sorpaas I 100% agree with you, _and i'm sure we are not the only 2_, on how *simply* editing a  final ECIP is against the 'protocol' and even comparable to an "irregular state change"  ( *puff* another unicorn dies* ) or the DAO. It's reverting/editing something that shouldn't be edited without going through the consensus protocol that is currently in place. 

Your second objection to this irregular state change (*puff* ... ) is that it has an hidden goal of trying to cover up for past actions of Afri. Seeing who his doing all this shenanigans, I tend to agree with you but one can never really tell what's in their head.

To that, and to all the **crazy people** here screaming 'doxxing' and tried to remove you from Owner with a shameless 'operation' the past 48h.. 

**The internet does not forget.**
**The internet does not forget.**
**The internet does not forget.**

If Afri and Isaac **REALLY** wish to be pseudo or anonymous, they need to atleast clean up their tracks otherwise it's just silly. They 'doxxed' themselves long time ago. You can't 'undoxx' someone. I'm surprised this is the best excuse you could find to try to kick @sorpaas out of the repo and community. 

**BUT** 

The proposed irregullar state change (*puff*) has zero relevance or impact for ETC..

and most importantly, the ETC values speak higher. 

If an individual asks for a simple thing like this, with zero impact on specifications or anything else, for the sake of **his/hers safety** (even if it does nothing for it or sounds like bullshit) we shouldn't be butting heads too long.

Let's either *kill another unicorn* or push an ammending ECIP as soon as possible, with ONLY this simple or removign their real names from it, instead of wasting more time on this.

But ya'll are crazy for thinking that this will save anyone from harm and this 'doxxing'-gate is plain bullshit coming from **'Team Krykoder'** ... yeah @TokenHash, that's who you are rolling with apparently, you sock puppet. Your lack of focus on the issue and wolfcrying for Team Krykoder will not be forgotten. You accused @sorpaas of personal attacks and other shit when you have ZERO moral ground. Look yourself in the mirror.

Fin.

PS: The Internet Does Not Forget.

sorpaas: Slightly off-topic, but indeed, it's kind of interesting to look back at those cover-ups https://web.archive.org/web/20200121172657/https://twitter.com/eth_classic/status/1103720719991607296

sorpaas:
Just a note and since this has been asked -- `soc1c` have been publicly using his real name "Afri" a lot of times. For example, in this meeting note of a call related to mining upgrades (https://github.com/ethereumclassic/ECIPs/issues/174#issuecomment-557106181). No one disclosed Afri's real name before he disclosed it himself. There's no basis why anything will ever be doxxing, unless you think Afri is doxxing himself.

That's why I'm opposing this. If you think about it, the whole thing is rather specious -- the doxxing argument comes right after security flaws are discovered in ECIP-1061, which Afri [played a major role](https://github.com/ethereumclassic/ECIPs/pull/220) in bypassing to process to push it forward. This makes people wonder whether this is an abuse of pseudonymity to escape responsibilities and to cover up wrongdoings.

PR #269, opened 2020-01-19
Progress ECIP-1048 (Clique) to Final Call.

bobsummerwill:
It was implemented a long time ago, but the ECIP was never updated.
We have to move it to Final Call first and then after is has been in that status, we can move it to Accepted, and then on to Final.
This is just house-keeping.


bobsummerwill:

PR #270, opened 2020-01-20
Updated ECIP-1019 to reflect that it was Replaced by ECIP-1017 and ECIP-1039

bobsummerwill:


bobsummerwill:

_specs/ecip-1019.md
```diff
@@ -3,14 +3,24 @@ lang: en
 ecip: 1019
 title: (Epoch Decay 10) Monetary Policy and Final Modification to the Ethereum Classic Emission Schedule
 author: Mike Boremi <ecip-1019@mikes.email>
-status: Draft
+status: Replaced
 type: Standards Track
 category: Core
 created: 2017-01-19
```
BelfordZ: Hey bob, great to see someone cleaning up shop! Thank you!

One thing ive noticed is that there is a meta data field for "replaces" that is suggested by the EIP template: https://github.com/ethereum/EIPs/blob/master/eip-template.md

That being said, do you think it would be more appropriate to use this field, and use the below NOTE as only an explanation of the surrounding circumstances?

bobsummerwill: Oops - we missed the boat, @BelfordZ.
I will make a new PR which adds the metadata.

bobsummerwill: Actually - I see that is still just a proposal for adding that additional metadata to EIPs.
And I think it is a decent enough proposal, but it isn't part of the EIP flow or the ECIP flow yet.
If you want to champion that, why don't you create an ECIP to propose it?
Or we can wait on the EIP parts to solidify and then adopt?

PR #278, opened 2020-01-27
ECIP-1081: No number in this ECIP

TokenHash:
Seems this ECIP was merged without a number. The actual underlying correction proposed was done in ECIP-1000 through another PR https://github.com/ethereumclassic/ECIPs/pull/267, so this ECIP may either get a number and moved to "Active" so it agrees with the ECIP process format, or it may be deleted.

### DIFF
--- a/_specs/ecip-1081.md
+++ b/_specs/ecip-1081.md
@@ -1,6 +1,6 @@
 ---
 lang: en
-ecip: **Please assign a number here and on the ECIP URL <-**
+ecip: 1081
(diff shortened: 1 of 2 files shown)


bobsummerwill:
Thanks, @TokenHash.

Yes, please can you assign a number, ECIP editors?  @meowsbits @soc1c @sorpaas @YazzyYaz @BelfordZ?

I will withdraw it in favor of ECIP-0001 when that is merged as Draft.

soc1c:
done

PR #280, opened 2020-01-28
ECIP-1061 "phoenix": aztlan as vanilla istanbul

soc1c:
open for discussion: a vanilla istanbul on classic.

benefits:
* no side effects or edge cases
* all clients support it out of the box
* we can stick to the original schedule
* phoenix day on march 25

### DIFF
--- a/_specs/ecip-1061.md
+++ b/_specs/ecip-1061.md
@@ -1,8 +1,9 @@
 ---
 lang: en
 ecip: 1061
-title: Aztlán EVM and Protocol Upgrades (Yingchun Edition)
-status: Accepted
+title: Aztlán EVM and Protocol Upgrades (Phoenix Edition)
+status: Last Call
+review-period-end: 2020-02-12
 type: Meta
 author: Talha Cross (@soc1c), Wei Tang (@sorpaas)
 created: 2019-06-06
@@ -12,11 +13,7 @@ discussions-to: https://github.com/ethereumclassic/ECIPs/issues/81
 ### Simple Summary
 
 Enable the outstanding Ethereum Foundation _Istanbul_ network protocol upgrades on the Ethereum
-Classic network without any gas-cost assumptions in a hard-fork code-named _Aztlán (Yingchun Edition)_ to enable 
-maximum compatibility across these networks. 
-
-The _Yingchun_ Edition refers to the preferred variant of _Aztlán_ without any repricing of state-trie 
-dependent opcodes as previously proposed in [ECIP-1072](https://ecips.ethereumclassic.org/ECIPs/ecip-1072).
+Classic network in a hard-fork code-named _Aztlán_ to enable maximum compatibility across these networks.
 
 ### Abstract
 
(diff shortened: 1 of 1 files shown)


bobsummerwill:
From ETC Discord:

That is a real terrible idea, @soc1c and @GregTheGreek.

The gas repricings in EIP-1884 break shipped smart contracts.  That lack of care about backwards compatibility is against the whole philosophy of ETC.   The removal of those changes from ETC was a very conscious and intentional decision.

We absolutely should not do that just to make an arbitrary date on the calendar.

We need to assess whether we can:

1. Make existing Aztlan + Phoenix timeline work.
2. Switch to Phoenix “redo” HF still hitting March 25th but with later activation dates for testnets
3. Switch to Phoenix “redo” but with later dates for all stages.
Yes, we are doing something different to ETH and that requires analysis (thank you Wei) and testing time.

We know that it can break contracts so we would not do it unless absolutely forced to by imminent threat.
It is just a line that we do not want to cross.

Ethereum compatibility is extremely desirable, but never the primary goal of ETC.

Also, the repricings (whether present or absent) do not make ETC more or less compatible.  But including them is known to break shipped smart contracts.

We need to be better.
If that means we need to delay then we must delay.
The scope was decided in that meeting and was very intentional.
I expect that ETC will diverge from ETH on gas pricing in the future, and that will be fine after Kunming is implemented.
ETC will likely lead the way on versioning.
And that will lead to (very slight) incompatibility again, but should be possible as “modes” within upstream tooling.
So impact to users is very minimal or unnoticeable.

Delaying would mean defining some new dates for the Phoenix HF ECIP:

http://ecips.ethereumclassic.org/ECIPs/ecip-1079

And coming to consensus on that as our new approach (and we would need a call to confirm that consensus).   Then ECIP-1061 and ECIP-1078 would be withdrawn while ECIP-1079 would move to Last Call.

And we should find a China-friendly platform (not Discord) for that call.  Also one which is easily recordable so we do not have the usual dicking around trying to get a recording to work on Discord.

TheEnthusiasticAs:
Skype works in China

bobsummerwill:
Also, not to be an arsehole here, but you yourself said that we could not take an Accepted ECIP backwards.   While that is perhaps a dubious assertion on @sorpaas’s account, I think that he has poisoned the possibility of revising 1061 (intentionally) by publishing https://github.com/multi-geth/multi-geth/releases/tag/v1.9.7 and with the corresponding Parity-Ethereum support for 1061 as it was on Dec 20th.

While we can argue whether or not “moving backwards” is acceptable I think it would be much better to work “well within the lines” of ECIP-1000 as written in this instance and so NOT to revise this ECIP.    Instead, a “HF which is 1061 with 1884 readded would be either an overlay HF (analogous to the Phoenix “fix”) or a new proposal (analogous to Phoenix “redo”).

bobsummerwill:
There are obviously no customers for those releases, because whatever @sorpaas asserts, nobody has any interest in shipping a “broken” fork, or “pushing” in a way which would compromise network security.

We just have a few different options and need to come to consensus on how best to handle the mis-specification which he highlighted (along with detailed proposals on how to “fix it”)

soc1c:
This is no longer viable with testnets activating it already.

PR #281, opened 2020-01-28
ECIP-1078: feature timing convention block number

meowsbits:
Feature-enable specifications conventionally use block numbers as inclusive
starting points, as opposed to last-without variables.

This commit edits the specification section assuming that that was what was
intended.


TheEnthusiasticAs:
I would lable this as "Aztlán Hardfork" project, nevertheless this project has already been merged.

PR #283, opened 2020-01-28
Withdraw ECIP-1081 Do Not Dox ECIP Editors.

bobsummerwill:
It was replaced by ECIP-0001 ECIP process.

### DIFF
--- a/_specs/ecip-1081.md
+++ b/_specs/ecip-1081.md
@@ -3,12 +3,18 @@ lang: en
 ecip: 1081
 title: Do Not Dox ECIP Editors
 author: Bob Summerwill (@bobsummerwill)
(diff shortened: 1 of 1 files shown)


TokenHash:
Yes, it was fixed on ECIP-1000 by PR https://github.com/ethereumclassic/ECIPs/pull/267, so 1081 needs to moved to `withdrawn`.

bobsummerwill:
@TokenHash Right!
I had lost track of the fact that this WAS actually already fixed in ECIP-1000.
So not even waiting for ECIP-0001.
Already fixed.  Cool.

bobsummerwill:
Thank you, @BelfordZ.

PR #286, opened 2020-02-02
ECIP-1085: Simple Subroutines for the EVM proposal.

bobsummerwill:


bobsummerwill:
Thanks for the 1085 assignment, @soc1c

PR #288, opened 2020-02-11
Move ECIP 1080: SELFBALANCE opcode to Last Draft.

bobsummerwill:
This should have happened at the same time as ECIP 1078: Phoenix EVM and Protocol Upgrades ("Aztlán fix") was moved.
Also correcting "type" and "category" from Meta to Standards Track.
That was a cut-and-paste error.


bobsummerwill:
Thanks, guys!

Issue #262, opened 2020-01-19
Phoenix Hard-fork to occur at the same block as Aztlan - to "fix it"


bobsummerwill:
Discussion for ECIP-1078 and ECIP-1086.

meowsbits:
Just a note that EIP1283 + EIP1706 =~ [EIP2200](https://github.com/ethereum/EIPs/pull/2200/files).

TheEnthusiasticAs:
I would lable this as "Aztlán Hardfork" project.

soc1c:
#293

meowsbits:
So as I understand it, this spec "allows" the incorrect EIP2200 specification as-is (SLOAD.constantGas=200), and will effectively do so on the Mordor and Kotti testnets for the span between the respective Aztlan and Phoenix activations. 

This prevents the EIP2200 fix (EIP2200.constantGas=800, eg. https://github.com/etclabscore/multi-geth/pull/144) from breaking the testnets by changing (actually correcting) the gas price without doing so in a hard fork. (Kind of a weird concept: the correction would break the incorrect network implementations; this proposal effectively delays the correction indefinitely on these affected networks).

meowsbits:
What if we were to also provide an activation number for the disabling of this allowance, especially with a value below the Phoenix-Fix (ECIP-1078) activation.

Pros:
- Allow us to test correct EIP2200 implementations and outcomes (both for logical soundness, and benchmarking live gas metering as it will be comparable to EIP1283+EIP1708).

Cons: 
- Aside from being a comparative reference, this configuration won't directly correspond to any Mainnet configuration, and could thus be seen as overkill.

soc1c:
> This prevents the EIP2200 fix (EIP2200.constantGas=800, eg. etclabscore/multi-geth#144) from breaking the testnets by changing (actually correcting) the gas price without doing so in a hard fork. (Kind of a weird concept: the correction would break the incorrect network implementations; this proposal effectively delays the correction indefinitely on these affected networks).

Exactly. Fun fact, such a thing used to happen on Ethereum Foundation mainnet and it was decided to rather specify the bug than rolling back the network.

bobsummerwill:
For the papertrail, some discussion from ETC Discord ...

Bob: ECIP-1086 makes a lot of sense to me and has my support.

soc1c: besu and geth already have PRs
not sure who would implement this in Parity

[...]

PR #293, opened 2020-02-18
ECIP-1086: SLOAD Gas Patch for the Testnets

soc1c:
Issues with the _Aztlán_ protocol upgrades were identified and sufficiently patched with the _Phoenix_ protocol upgrades. However, the subsequent activation of these upgrades on the _Kotti_ and _Mordor_ testnets revealed inconsistencies between client implementations. This document proposes a patch for the aforementioned testnets.

### Abstract
The [EIP-2200](https://eips.ethereum.org/EIPS/eip-2200) defines structured definitions of net-gas metering for the `SSTORE` opcode. This EIP was activated in a bundle called [EIP-1679](https://eips.ethereum.org/EIPS/eip-1679) "Istanbul" on the Ethereum Foundation network and [ECIP-1061](https://ecips.ethereumclassic.org/ECIPs/ecip-1061) "Aztlán" on the Ethereum Classic network.

Notably, the difference here is that _Istanbul_ contained [EIP-1884](https://eips.ethereum.org/EIPS/eip-1884) which repriced various opcodes for trie-size dependent operations, whereas _Aztlán_ did not.

Both [EIP-1884](https://eips.ethereum.org/EIPS/eip-1884) and [EIP-2200](https://eips.ethereum.org/EIPS/eip-2200) propose the `SLOAD_GAS` to be increased from `200` to `800` but not all clients correctly implemented this change for both EIPs leaving networks configured with a pick-and-mix EIP configuration such as _Aztlán_ with an incompatible configuration as compared to clients who correctly implemented the specifications.

### Motivation
This specification seeks to validate the incomplete configuration in favor of retaining the _Kotti_ and _Mordor_ testnets in consensus. This ECIP shall not be considered on mainnet. It's sole purpose is to allow avoiding a rollback on the testnets and having a proper documentation on the cause for the gas mismatch and providing instructions for client developers how to circumvent this.

### Discussions to:
- https://github.com/ethereumclassic/ECIPs/issues/262

### DIFF
--- a/_specs/ecip-1086.md
+++ b/_specs/ecip-1086.md
@@ -60,6 +60,8 @@ Note, that accepting and merging these pull requests would either require to rol
 
 Adoption of the content of this ECIP requires a configuration for all clients to allow for configuring `SLOAD_GAS` as per [ECIP-1086](https://ecips.ethereumclassic.org/ECIPs/ecip-1086). The implementation cost is considered low.
 
+A sample implementation is available for _Besu_: [hyperledger/besu#382](https://github.com/hyperledger/besu/pull/382)
+
 ## Copyright
 
 Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/_specs/ecip-1086.md
+++ b/_specs/ecip-1086.md
@@ -60,7 +60,11 @@ Note, that accepting and merging these pull requests would either require to rol
 
 Adoption of the content of this ECIP requires a configuration for all clients to allow for configuring `SLOAD_GAS` as per [ECIP-1086](https://ecips.ethereumclassic.org/ECIPs/ecip-1086). The implementation cost is considered low.
 
-A sample implementation is available for _Besu_: [hyperledger/besu#382](https://github.com/hyperledger/besu/pull/382)
+An implementation is available for:
+
+* Hyperledger Besu: [hyperledger/besu#382](https://github.com/hyperledger/besu/pull/382)
+* Parity Ethereum: `N/A`
+* Multi-Geth: [etclabscore/multi-geth#151](https://github.com/etclabscore/multi-geth/pull/151)
 
 ## Copyright
 
--- a/_specs/ecip-1086.md
+++ b/_specs/ecip-1086.md
@@ -1,0 +1,65 @@
+---
+lang: en
+ecip: 1086
+title: SLOAD Gas Patch for the Classic Testnets
+status: Draft
+type: Standards Track
+category: Core
+author: Talha Cross <no-reply@soc1.cz>
+created: 2020-02-18
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/262
+---
+
+### Simple Summary
+
+Issues with the _Aztlán_ protocol upgrades were identified and sufficiently patched with the _Phoenix_ protocol upgrades. However, the subsequent activation of these upgrades on the _Kotti_ and _Mordor_ testnets revealed inconsistencies between client implementations. This document proposes a patch for the aforementioned testnets.
+
+### Abstract
+
+The [EIP-2200](https://eips.ethereum.org/EIPS/eip-2200) defines structured definitions of net-gas metering for the `SSTORE` opcode. This EIP was activated in a bundle called [EIP-1679](https://eips.ethereum.org/EIPS/eip-1679) "Istanbul" on the Ethereum Foundation network and [ECIP-1061](https://ecips.ethereumclassic.org/ECIPs/ecip-1061) "Aztlán" on the Ethereum Classic network.
+
+Notably, the difference here is that _Istanbul_ contained [EIP-1884](https://eips.ethereum.org/EIPS/eip-1884) which repriced various opcodes for trie-size dependent operations, whereas _Aztlán_ did not.
+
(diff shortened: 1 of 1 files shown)


soc1c:
cc @GregTheGreek @edwardmack

cc @meowsbits @tzdybal

bobsummerwill:
LGTM

bobsummerwill:
For the papertrail, some discussion from ETC Discord ...

Bob: ECIP-1086 makes a lot of sense to me and has my support.

soc1c: besu and geth already have PRs
not sure who would implement this in Parity

Bob: @soc1c So that would likely have to be ChainSafe.  I am happy for ETC Cooperative to pay for that work, if there is engineering capacity available for that, @GregTheGreek?

GregTheGreek: @bobsummerwill We can definitely do it

Donald: seems the name should be changed to "ChainSave" 

Cody: We control the validators on kotti seems straight forward. There shouldn't be any expectations that testnets wont be reset periodically.
I agree ecip1086 solves the problem, but it is a testnet problem. We found a problem with an implementation and know how to fix it. We can just pick a block number and redo
Resetting the testnet sets expectations

Bob: And that is the plan, right, @DontPanicBurns?
And ECIP-1086 just the specification of that plan to coordinate that switchover between the different clients.
We can argue that there is no need for the ECIP - that we can just agree to make those changes and switch over - just by the clients making that change themselves - but the specification gives us the papertrail and coordination.
I don't see any downside to pushing that ECIP through the process as well as doing the work within the clients.
We do include testnet activation blocks within the ECIP process too, which you could argue does not need that same degree of coordination.
But we have all the same coordination problems with the shared, public testnets as we do with the mainnet.   Just not the same economic impact.

Cody:  That is not the plan as proposed in 1086
We are doing a testnet only fork. This will not be the same fork that will happen on mainnet

Bob: Right.

Cody: It needs to be a clean test that mirrors exactly what will happen on mainnet when aztlan happens
We are not doing a one two punch on mainnet, we should not do a one two punch on testnet

Bob: So that test is not possible on Mordor and Kotti anymore.
ECIP-1086 is about getting those back in consensus so they can be used.

To do what you are suggesting would need additional new testnets to be created.
As happened with Nazgul for Atlantis.

Cody: Why is it not possible

Bob: And maybe that should happen too, yes.
"Why is it not possible"

Not possible without resetting which might screw end-users up in a way which we cannot quantify.

Cody:  It is a testnet
We are the enduser

Bob: But "we" (ie. the group talking here) are not all the users of Kotti.

Cody:  Expectation set
Everything on the testnet can be trivially resubmitted
If it isnt trivial, it should be running on mainnet for a fee
We dont know what happens when all the eips/ecips are activated at the same time

Bob: You are describing a desirable world, but that world is not the world we are currently living in.   We do not and cannot know the consequences onto end users of resetting the testnets.

There was no clear expectation set that testnets may be reset at will.

"We dont know what happens when all the eips/ecips are activated at the same time"

Sounds like we need some more (temporary) testnets as well as Kotti and Mordor.

Cody: No we need to reset the screed up fork in the yestnet
Or throw the other two in the garbage

Bob: Which would do the atomic Aztlan+Phoenix activation at the same block.
We can dry run that transition multiple times on those testnets.

Same as for the Keccak256 tests.   That Ethash to Keccak256 transition is something which you want to dry-run (for all of the stack including clients, mining software, exchanges, wallets) multiple times.

Cody: Kotti is that testnet that should be restarted frequently.

Bob: Maybe.

But that is a change of expectations which would need to be worked through the ECIP process itself, build human consensus, etc.

Cody: Or make an un fork block
O done see any change
Is there an ecip creating the kotti network?
Or anything stating a general view of "why testnets exist"
4 people need to update their nodes to change anything on kotti
Or maybe 5 I forget how many validator we have
Its equally a pain in my ass to update my node to reset the testnet as it is to update my node to do forks

Bob: "Is there an ecip creating the kotti network?  Or anything stating a general view of "why testnets exist"

There should be.
Because at some point it became an implicit assumption that any ETC mainnet consensus change should first make its way through the ETC testnets.   First, I guess, only assuming Classic Morden, at some point adding Kotti to the mix, and then with Agharta including Mordor instead of Classic Morden.

There was an ECIP for defining Clique, but the creation of Gorli and Kotti happened outside of the ECIP process.   @soc1c would be more aware of that history than me.
ECIP-1000 and eventually ECIP-0001 should probably define our expectations about testnets, and our specific expectations for Mordor and Kotti during the phase where we are "dry-running" hardforks.

Even if we have not explicitly specified it as a requirement, I think it exceedingly unlikely that any HF would be Accepted if it did not include activation blocks for both Mordor and Kotti, with the expectation that the ECIPs included in that HF would be tested on both testnets prior to mainnet activation.
@GregTheGreek @isaac @soc1c Would be exceedingly useful for you guys to speak up at this stage, either here on the Discord or better yet on the Github issues and PRs.
Or better yet, @DontPanicBurns, for you and me to take this discussion  there too.

Greg:  Theres honestly too much noise at this point
its becoming increasignly difficult to monitor and help this channel
I believe i approved @soc1c pr, it logically makes sense
To respond to @DontPanicBurns i completely disagree with you. Rolling back a testnet is not the right way to go about things. Patching all the way. Treat your testnets like its your mainnet.
HF dev is the same as traditional software dev. Your stages:
Dev -> Staging -> Prod

In blockchain world how isee it:
Local or personal networks -> Testnets -> Mainnet
Traditionally speaking, if you broke staging and had to revert, its a failure.

Husain: only 4 different people other than you (Bob/Tala/Cody/Donald ) have been contributing since the 6th of Feb...
and only in the last few days has this been active with Donald/Bob discussing ECIP0001 and now the testnet roll back issue
how on earth is it becoming increasingly difficult to monitor this channel?

Cody:  Traditionally speaking, if you broke staging and had to revert, its a failure.

Staging was broke it revealed the error, we need to revert
I approved the ecip because it is proper ally formated

YazzyYaz:
LGTM, merging

PR #295, opened 2020-02-20
ECIP1087 [Draft]: A way to resolve the ECIP1061/ECIP1078/ECIP1086/EIP2200 testnet fiasco (rollback Mordor, leave Kotti)

meowsbits:
### Abstract

Don't touch Kotti; allow ECIP1061, ECIP1078, and ECIP1086 (EIP2200, EIP2200Disable) to continue as specified and implemented there.

Set Mordor ECIP1078 activation to 778507 (equivalent to ECIP1061), and rollback to 750000.

### Rationale

Kotti is a PoA network, so it doesn't simulate (ie test) ETC mainnet in the first place (at least not well) since it uses different consensus rules. If it is a public simulation and test of ETC mainnet, it isn't a good one. So we can follow the logic of A: Testnets-as-service.

Mordor is a PoW network, so is is a pretty good simulator of ETC mainnet. We can follow the logic of B: Testnets-as-test, and not only resolve the issue of EIP2200 correction, but improve the network's simulation of mainnet activation for ECIP1061 and ECIP1078.

---

Rendered: https://github.com/meowsbits/ECIPs/blob/feat/ecip1087/_specs/ecip-1087.md

### DIFF
--- a/_specs/ecip-1087.md
+++ b/_specs/ecip-1087.md
@@ -12,7 +12,7 @@ license: Apache-2.0
 
 Don't touch Kotti; allow ECIP1061, ECIP1078, and ECIP1086 (EIP2200, EIP2200Disable) to continue as specified and implemented there.
 
-Rollback Mordor to 750000 (3507 blocks prior to ECIP1061 activation), with ECIP1078 activation at block 778507 (same as ECIP1061).
+Set Mordor ECIP1078 activation to 778507 (equivalent to ECIP1061), and rollback to 750000.
 
 ### Motivation
 
--- a/_specs/ecip-1087.md
+++ b/_specs/ecip-1087.md
@@ -12,7 +12,7 @@ license: Apache-2.0
 
 Don't touch Kotti; allow ECIP1061, ECIP1078, and ECIP1086 (EIP2200, EIP2200Disable) to continue as specified and implemented there.
 
-Rollback Mordor to 750000, activate ECIP1078 at block 778508.
+Rollback Mordor to 750000, activate ECIP1078 at block 778507.
 
 ### Motivation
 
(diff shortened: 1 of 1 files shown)


soc1c:
I agree with the rationale. Applying 1086 to Kotti makes a lot of sense.

To simplify the testnet reset on Mordor, I would propose to move both Aztlan and Phoenix to block `976_231`. That way, we would invalidate the current chain anyways and force all clients to roll back.

soc1c:
I actually don't think we would need an ECIP for that. See #296 directly applying your suggestion.

soc1c:
would favor #297 instead and have a new specification for a "good" phoenix. see #298

YazzyYaz:
Should this be closed now?

meowsbits:
@YazzyYaz How about I'll edit to `Withdrawn` and we can merge for the sake for documentation (along w/ 1089).

TokenHash:
i approved "withdrawn" status

pls second approval and move to merge

meowsbits:
@ethereumclassic/ecip-editors please review for editorial righteousness for the doc-only merge

TokenHash:
@meowsbits License is clarified at the header, but I think if you add these details at the bottom, it's fine editorially:

### Implementation

N/A

### Copyright/Licensing

This work is licensed under [Apache License, Version
2.0](http://www.apache.org/licenses/).

bobsummerwill:
Is this obsoleted by Phoenix, @meowsbits?
Maybe just Close it now?

meowsbits:
Yea it'll probably be practically obsoleted, but I think it should be merged because
- https://github.com/ethereumclassic/ECIPs/pull/293/files was, which would by the same reasoning be considered obsoleted and this doc addresses a related issue
- it has a strong rationale for differentiating Mordor and Kotti (as PoW and  PoA) network utilities and management priorities (ie testnet as test vs. service)
- it has a number 1087 which would be left missing in the list which would bother me

TokenHash:
@bobsummerwill @meowsbits ...yes, I think it went thru a cycle and has to be merged as it was proposed, draft, and withdrawn...actually it should not go thru these cycles as a PR! ;)

meowsbits:
@ethereumclassic/ecip-editors beep boop Now we've established at least 3 times that this is for documentation only... please review for the format for the merge for the the PR-inbox sanity.

PR #296, opened 2020-02-21
ECIP-1061: Move Aztlan to Phoenix block on Mordor; ECIP-1086 Last Call

soc1c:
The rationale is taken from #295 

* revert mordor and have a proper phoenix + aztlan hardfork
* keep kotti with ecip-1086 #293 

cc @meowsbits

### DIFF
--- a/_specs/ecip-1061.md
+++ b/_specs/ecip-1061.md
@@ -31,9 +31,9 @@ _Istanbul_ hardforks. The proposed changes for Ethereum Classic's _Aztlán_ upgr
 
 This document proposes the following blocks at which to implement these changes in the Classic networks:
 
-- `778_507` on Mordor Classic PoW-testnet (activated on Jan 30th, 2020)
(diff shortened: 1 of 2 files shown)


soc1c:
@meowsbits no. because the changes proposed are not standalone. this should be accepted as bundle or rejected as bundle. but you wouldn't be able to pick from the two change sets.

meowsbits:
With my latest understanding of EIP2200 `SLOAD` gas specification (and related patches), it seems to me that ECIP1086 is invalid/useless, since because EIP2200 does _not_ specify `SLOAD` gas=800, no patch should be implemented (or now, should actually be reverted in some client cases), and ECIP1086 as a "retrospective allowance" for the EIP2200 gas=200 become a no-op.

meowsbits:
Would the Mordor revert reuse the same genesis, or a new one to prevent forks?

soc1c:
Closing in favor of #297

PR #297, opened 2020-02-21
reject aztlan and pull phoenix

soc1c:
due to the multiple issues disclosed with aztlan and phoenix not being a sufficient fix and both testnets being broken, this pull request proposes to reject ecips-1061, 1078, 1086 due to security and stability concerns and to move on with writing a phoenix proposal from scratch. 

accepting this pull request requires community consensus in a publicly announced call.


GregTheGreek:
The alternative is to implement instanbul outright, correct?

PR #302, opened 2020-02-26
fix(Gemfile.lock): bundle update and fix nokogiri security alert

meowsbits:
Should resolve reported vulnerability with nokogiri < 1.10.8

Signed-off-by: meows <b5c6@protonmail.com>


meowsbits:
Closing in favor of a simpler and more comprehensive feature and fix bundle at #306 , which includes this change and then makes more things better too.

If anyone wants to reopen to make the change set more modular I'm happy to do so. Just trying to save a few editor-review bandwidth.

PR #304, opened 2020-02-27
fix(ecip1082): character that cannot start any token, line 5, col 9

meowsbits:
- broken: https://github.com/meowsbits/ECIPs/runs/470695379?check_suite_focus=true
- fixed: https://github.com/meowsbits/ECIPs/runs/472177250?check_suite_focus=true

### DIFF
--- a/_specs/ecip-1082.md
+++ b/_specs/ecip-1082.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1082
 title: Fork identifier (`forkid`) protocol fork list specification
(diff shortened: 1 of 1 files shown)


meowsbits:
Closing in favor of a simpler and more comprehensive feature and fix bundle at #306 , which includes this change and then makes more things better too.

If anyone wants to reopen to make the change set more modular I'm happy to do so. Just trying to save a few editor-review bandwidth.

PR #305, opened 2020-02-27
CI update for ECIP validation

meowsbits:
(Branch is based on and includes changes from #302's nokogiri vulnerability fix)

I made a fork of the `eip_validator` tool at https://github.com/meowsbits/ecip_validator that handles ECIP frontmatter validation. Happy to transfer that project to ethereumclassic if there's interest in continuing development there, just let me know.

Anyways this change set
- revises workflow names to separate Jekyll build from the ECIP validation 
- creates ECIP validation workflow.

You can see what these look like at https://github.com/meowsbits/ECIPs/actions

Currently the ECIP validation is failing because quite a few specs aren't up to snuff.  Fixing those can be a later job.

PR #306, opened 2020-02-27
ECIP validation fixes and CI improvements

### LINKED ISSUE #292: web view broken for (some?) ECIPs - frontmatter
Seems to be an issue with frontmatter. eg. for ECIP-1082

Should this validation be handled on CI?

---

https://ecips.ethereumclassic.org/ECIPs/ecip-1082 

2020-02-17-115239_842x959_screenshot

meowsbits:
Builds from #305 

Fixes all ECIP document validation issues and gets the CI suite to pass (and includes a new workflow `ECIP Validation` which validates ECIP document formats on `pull_request` and `push`). So we can depend on CI validation and deploy workflow as a requirement for merging PRs, for example.

Includes change sets:
- fix(Gemfile.lock): bundle update and fix nokogiri security alert #302
- fix(ecip1082): character that cannot start any token, line 5, col 9 #304
- CI update for ECIP validation #305 

Fixes #292 

As you'll see in those individual PRs I've noted that this PR includes and then extends from them.
If anyone wants to reopen those more granular changes please just say so and we can do that. Fixing the ECIP documents for the validator to pass meant tweaking and fixing frontmatter fields and values, but otherwise no actual content was edited, and I tried to mess as little with the frontmatters as possible, eg.
- https://github.com/ethereumclassic/ECIPs/pull/306/commits/dc20abfe315acd1c6213f686b643101824e821cf
- https://github.com/ethereumclassic/ECIPs/pull/306/commits/431d7f8b6cfebb7fc1dba8c117e240db94368ac4
- https://github.com/ethereumclassic/ECIPs/pull/306/commits/7a6a9a517fda9254422269fa957837aad62ac133

The new validator can be found here: https://github.com/meowsbits/ecip_validator and is published with rubygems here https://rubygems.org/gems/ecip_validator. As mentioned in #305, if there's interest in pushing forward a community-development version, I'd be happy to transfer it to github.com/ethereumclassic/.


meowsbits:
Passing workflows visible here: https://github.com/ethereumclassic/ECIPs/actions

_specs/ecip-1015.md
```diff
@@ -6,6 +6,7 @@ author: Cody W Burns <cody.w.burns@ethereumclassic.org>
 status: Final
 type: Standards Track
 category: Core
+created: 2016-10-25
```
meowsbits: Source: https://github.com/ethereumproject/ECIPs/commit/925a3e87babf211a9df210fa3a84344742d00ba7

meowsbits:
@ethereumclassic/ecip-editors ping for the review plz

Gemfile.lock
```diff
@@ -210,60 +208,65 @@ GEM
...
-    yell (2.0.7)
+    unicode-display_width (1.6.1)
+    yell (2.2.2)
 
 PLATFORMS
   ruby
 
 DEPENDENCIES
-  eip_validator (>= 0.8.0)
+  ecip_validator (>= 0.9.0)
```
BelfordZ: Not the biggest fan of this being an external dep.

since a tool that validates ECIPs is kind of part of the ECIP spec itself. Who maintains this tool? Where is it located?

.github/workflows/jekyll.yml
```diff
@@ -10,6 +10,7 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Build the site in the jekyll/builder container
+      if: github.repository == 'ethereumclassic/ECIPs'
```
BelfordZ: why would it be anything else?

Gemfile.lock
meowsbits: > The new validator can be found here: https://github.com/meowsbits/ecip_validator and is published with rubygems here https://rubygems.org/gems/ecip_validator. As mentioned in #305, if there's interest in pushing forward a community-development version, I'd be happy to transfer it to github.com/ethereumclassic/.

.github/workflows/jekyll.yml
meowsbits: Forks?

Gemfile.lock
meowsbits: Yea, I can readily get on board with moving it closer to this repo... even into this repo. But I'm just thinking one step at a time, at least to use it as-is to get the ECIPs straightened out, since this is how EIPs do it, so it's an established pattern, and if there's energy to restructure/replace/relocate the tool then that's great but doesn't seem like a dependency for the spirit of this PR to me.

meowsbits: And I do really think that would be something worth prioritizing here too, since that code will be just another representation of some of the parameters outlined in ECIP1000 or whatever, like requiring frontmatter fields, values, whatever. Actually even a better way of doing it than ECIP1000/whatever imo.

PR #307, opened 2020-03-04
Update ecip-1037.md

realcodywburns:
format clean up


bobsummerwill:
@q9f @YazzyYaz @BelfordZ  Looks like this is just waiting on trivial approval?

PR #310, opened 2020-03-06
ECIP-0001: Fixed url capitalization - added language to include controversies

TokenHash:
• ECIP -> ecip

• Added language "including, but not limited to" in the phrase that clarifies that devs/editors can change ECIPs to previous statuses in the case in the examples but also if other major events occur, such as the recent ProgPow controversy after EIP-1057 was moved to accepted. The text modified was:

"However, core developers and editors can decide to move ECIPs back to previous statuses after the fact, or to `Rejected` or `Withdrawn` status, at their discretion provided there is reasonable cause, **including but not limited to,** a major, but correctable, flaw was found in the ECIP, a major but not correctable flaw was found in the ECIP."

### DIFF
--- a/_specs/ecip-0001.md
+++ b/_specs/ecip-0001.md
@@ -203,7 +203,7 @@ For Meta or Informational ECIPs:
 WIP -> DRAFT -> LAST CALL -> ACTIVE
 ````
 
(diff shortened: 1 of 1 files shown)


bobsummerwill:
LGTM

PR #291, opened 2020-02-17
ECIP-1082: Last Call

meowsbits:
Original document here: https://github.com/ethereumclassic/ECIPs/blob/master/_specs/ecip-1082.md
Revised document rendered: https://github.com/ethereumclassic/ECIPs/blob/ad181baeb6361873d5e77bd3eaa945d63b1ad904/_specs/ecip-1082.md

PR #311, opened 2020-03-11
_specs: 1082 status=Final

meowsbits:
It has been at least 2 weeks since this doc has been at
Last Call (https://github.com/ethereumclassic/ECIPs/pull/291).

Signed-off-by: meows <b5c6@protonmail.com>


meowsbits:
@ethereumclassic/ecip-editors beep for the review for the merge for the Final

Since we've got some more forks since this was drafted, I'll plan to follow up with additional `Implementation` examples once this gets Final.

meowsbits:
Noting w/r/t [the above `mentioned` link](https://github.com/multi-geth/multi-geth/pull/141): It's a PR to multi-geth/multi-geth with test cases showing that that provider hasn't yet implemented the protocol to spec, yet.

Not yet sure about hyperledger/besu or others yet. But if multi-geth/multi-geth has it wrong, then others might as well.

bobsummerwill:
@meowsbits This is related to https://eips.ethereum.org/EIPS/eip-2364, right?
And https://eips.ethereum.org/EIPS/eip-2124?

Both still in Draft status but implemented in upstream Geth?   And also not affecting consensus, so not needing a HF to implement.

Are you are seeking to implement this within Core-Geth so that the Geth <-> Core-Geth discovery and uncoupling is cleaner/faster?

bobsummerwill:
PS.  Would it be helpful for the ETC fork metadata in Core-Geth to up upstreamed to Geth as well, or does it not make a difference?

meowsbits:
> This is related to https://eips.ethereum.org/EIPS/eip-2364, right?
And https://eips.ethereum.org/EIPS/eip-2124?

Yep.  https://github.com/ethereumclassic/ECIPs/blob/b1c48b16d58de8ac6b8a159eef66901d9d80c8af/_specs/ecip-1082.md#abstract

> Are you are seeking to implement this within Core-Geth so that the Geth <-> Core-Geth discovery and uncoupling is cleaner/faster?

It is already implemented. Please review for the original EIP documents for rationale.

> Would it be helpful for the ETC fork metadata in Core-Geth to up upstreamed to Geth as well, or does it not make a difference?

`forkid` is a function of chain configuration. Thus, you can't "upstream metadata" without upstream the chain config, which ethereum/go-ethereum is reluctant to accept at this point.

bobsummerwill:
Thanks for the clarifications.

BelfordZ:
@meowsbits just needs to pass validation.

realcodywburns:
Failing with:

_specs/ecip-1083.md is NOT valid:	 {:lang=>["can't be blank"]}
_specs/ecip-1087.md is NOT valid:	 {:lang=>["can't be blank"], :author=>["can't be blank"], :type=>["is not included in the list"]}

meowsbits:
:green_circle: Passed: https://github.com/ethereumclassic/ECIPs/runs/727058198
:purple_circle: Merged.

PR #312, opened 2020-03-18
Integrating ecip-0001 into ecip-1000

TokenHash:
To make it easier to compare and move to active, I am integrating the changes I proposed in ecip-0001 into this ecip-1000:

• it leaves Sorpaas as editor
• it keeps the changes and additions to rules proposed in ecip-0001

In summary, the new text largely clarifies some aspects of ecip-1000 that were not explicit before.

The goal is to minimize discussions about "process" and focus on standards.

### DIFF
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -5,7 +5,7 @@ title: ECIP Process
 status: Active
 type: Meta
 discussions-to: https://github.com/ethereumclassic/ECIPs/issues/58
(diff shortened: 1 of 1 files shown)


gitr0n1n:
"# Precedents"

Might be worth the attributions: isn't Bitcoin's BIPs based off of Python's PEPS?
https://github.com/python/peps

gitr0n1n:
Line 384: "ECIP-0001" reference to old ECIP. @TokenHash

PR #326, opened 2020-07-16
MV Sorpaas from current to former ecip editor

realcodywburns:
Moving Wei Tang (@sorpaas) from "current editors" to "former editors" to comply with their wish to disengage from Ethereum classic ecosystem. Reference: https://that.world/~classic/2020/06/10/deprecate/


bobsummerwill:
@realcodywburns Are there Github permissions which need updating too?

realcodywburns:
Yes I will move them to outside contributor as well

Issue #327, opened 2020-08-07
ECIP-1092 51attack solution: PirlGuard & Callisto proposal


Dexaran:
---
lang: en
ecip: 1092
title: 51% attack solution: PirlGuard & Callisto
status: Draft
type: Core
author: dexaran (@dexaran)
created: 8/7/2020
license: LGPLv3
---

# Abstract

The following describes a method of preventing a 51% attack on Ethash-based POW chains.

# Motivation

Ethereum CLassic suffered a number of 51% attacks already. [Here](https://cointelegraph.com/news/is-etc-102-screwed-after-second-51-attack) you can find an article from Cointelegraph describing the 51% attack on 6 Aug 2020. [Here](https://cointelegraph.com/news/ethereum-classic-51-attack-the-reality-of-proof-of-work) you can find another article describing the 51% attack on 10 Jan 2019. As long as ETC protocol remains unchanged it is susceptible for more 51% attacks.

# Specification

The proposed solution refers to the [PirlGuard protocol](https://pirl.io/en/pirlguard-innovative-solution-against-51-attacks/). 

The description of the protocol states that instead of automatically syncing with any privately-mined chain branch the new protocol should require the peer proposing the privately-chain (and the reversal of the publicly mined and synced blocks) to mine  a number of "penalty" blocks. The number of penalty blocks must depend on the number of the original blocks that would be reverted if the chain will reorganize and sync to the proposed (privately-mined) branch. Thus the cost of the 51% attack will dramatically increase as the attacker will not be able to mine his private branch and then broadcast it to the network thus reverting all the transactions from the "main" branch.

# Rationale

This protocol is already implemented in Pirl and Callisto Network. The protocol has a working time tested reference implementation. The proposed change requires minimal modifications of the Ethereum CLassic protocol while offers a reliable method of protecting against the recent attacks and the relevant threat of newer similar attacks.

# Implementations

Here is a reference implementation of the protocol in GO lang: https://github.com/EthereumCommonwealth/go-callisto/blob/master/core/pirl_guard.go

https://github.com/EthereumCommonwealth/go-callisto/blob/master/core/blockchain.go#L1540

# Presentation

Made by @padenmoss

http://padenmoss.com/public/ECIP-1092_proposal.pdf

# Copyright

This ECIP is licensed under [GNU Lesser General Public License v3.0](https://www.gnu.org/licenses/lgpl-3.0.en.html).

BaikalMine:
Its a good idea!

[...]

PR #328, opened 2020-08-07
ECIP-1092 51% attack solution: PirlGuard & Callisto implementation

Dexaran:
Proposal text is a copy of [this corresponding issue](https://github.com/ethereumclassic/ECIPs/issues/327).


TheEnthusiasticAs:
From several days of discussions of this topic in the ETC discord, I summarized the pro and contra arguments, what I could collect:

Pros:
- a working solution on the Callisto Network
- Can be implemented like a "copy-paste"
- could be seen as a temporary solution till a better one is found

Contras:
- some developers have technical doubts till they see security report(s)
- it makes possible to have a chain with less PoW to win (PoW rule: a chain with the most PoW wins)
   -> violation of the censorship-resistance-principle (because one mined offline it will be penalized)
- some other blockchain principles would be broken, what one can't see at them moment
- an attacker is e.g. making 3k+ block reorgs. They have the time and money to check in and then mine 3k more honestly

Please extend/correct, if you have any suggestions!

_specs/ecip-51attack-solution.md
```diff
@@ -0,0 +1,39 @@
+---
+lang: en
+ecip: ?
```
realcodywburns, line 3: please add number 1092

Dexaran: done

PR #339, opened 2020-08-17
Update ECBP 1076 "Miner Signaling" with latest options

q9f:
Incorporates reviews from #194 and adds RandomX


developerkevin:
Is "miner signaling" non-binding or do we take what is being signaled as bond i.e. miners will decide X or Y and will be added or change implemented?

_(Just trying to gain some clarity)_

q9f:
This needs some more thoughts. I do not wish to proceed with that currently.

PR #367, opened 2020-09-10
ECIP 1099: Calibrate Epoch Duration

iquidus:
Add ECIP 1099 "Calibrate Epoch Duration" draft.


q9f:
Thanks for the submission. As this is an alternative proposal to 1043, I would put it on the agenda for tomorrows call. Would you be able to attend?

q9f (approved):

realcodywburns (approved):
lgtm

Issue #368, opened 2020-09-10
ECIP-1099 Calibrate Epoch Duration


iquidus:
This issue is the discussion for #367 which proposes extending epoch duration.

q9f:
Thanks for proposing this. My first thoughts:

Pro: 
* :+1: it's a good temporary 51% attack solution as it modifies ethash enough to invalidate hashrate available on nicehash et al.
* :+1: we kind of missed the activation window for ECIP-1043 DAG size limitation (epoch < 350), this is a viable solution even right now or in 2-3 months
* :+1: this will not only maintain long-term compatibility with current 4GB mining cards but also re-enable 2GB and 3GB GPU mining for a while

Con:
*  :warning: needs a hardfork
*  :warning: needs custom miner software

q9f:
moved to last call and replaces 1043. #370 #371

as per decision at call #362 (13)

Lolliedieb:
With the fork happening after 4G cards are already ruled out on epoch 383+ I wonder how remaining miners with 6G or 8G cards will react. Usually people do not choose their own butcher - I mean here in terms of the hard hit of profitability the network will take if many 4G cards (and 3G as well) will get back active. The danger is here that 8G miners vote against the consensus change and mine on a not updated node causing a chain split. 

Also I doubt that this will help much against 50% attacks, because when eth also has ruled out 4G cards likely many cards will turn of (the etc network can only be profitable for a very limited fraction of the sub 4G miners). The price determines the hashrate, not the other way around - at least as long as overall enough mining GPUs exist that could switch the network.  That said I expect only a few weeks after the action the effect of this change would be gone and the network would be attack-able as before because drawing hashrate from eth to etc would be too easy. 

I would rather push the efforts towards creating a new proof of work that is different enough from ethash.

sjococo:
I object to any changes to the dag file creation by for example as it is proposed now. In the past no change was made to let 1 GB cards mine, no change was made to let the 2 GB cards mine, no change was made to let the 3 GB cards mine. 

So now no change can be made to let the 4 GB cards mine after the DAG file size exceeds their on board memory. It is an attack to the ETC blockchain itself what grew organic and has to be maintained as it was intended.

If you want to change it, you should hard fork to a new coin and let ETC evolve and continue as it did.

Ordfin:
[...]

Issue #362, opened 2020-09-03
ETC Core Devs Call 13 & 14


q9f:
# ETC Core Devs Call 14
* When: **Friday, September 25, 2020, 4pm UTC,** 90 minutes max
* Where: [ETC Discord](https://discord.gg/x3WHgrU)
* Focus: Mining Algorithm, Last-Call Proposals, and other 51%-Attack Solutions.

### Agenda

* Last-Call Proposals
    * ~~ECIP 1049 KECCAK256 (Last Call)~~ Moved to call 15 on request
    * [ECIP 1099](https://ecips.ethereumclassic.org/ECIPs/ecip-1099) Calibrate Epoch Duration (Last Call)
* 51% Proposals
  * [ECIP 1092](https://ecips.ethereumclassic.org/ECIPs/ecip-1092) 51-percent attack solution PirlGuard by Callisto (Draft)
  * [ECIP 1094](https://ecips.ethereumclassic.org/ECIPs/ecip-1094) VeriBlock Proof-of-Proof 51%-Attack Prevention (Draft)
  * [ECIP 1096](https://ecips.ethereumclassic.org/ECIPs/ecip-1096) 51% Attack protection system based on Bitcoin Merged Mining (Draft)
  * [ECIP 1097](https://ecips.ethereumclassic.org/ECIPs/ecip-1097) Checkpointing based 51% attack resistance (Draft)
  * [ECIP 1100](https://ecips.ethereumclassic.org/ECIPs/ecip-1100) Modified Exponential Subjective Scoring (Draft)

Live-Agenda: https://hackmd.io/@q9/etc-dev-14

~~_Please post all relevant proposals below that should be discussed and evaluated._~~ Agenda submissions are closed.

# ETC Core Devs Call 13
* When: **Friday, September 11, 2020, 4pm UTC,** 90 minutes max
* Where: Jitsi 
  * Web/App [ETCCoreDevsCall13](https://meet.jit.si/ETCCoreDevsCall13#config.startWithVideoMuted=true&config.startVideoMuted=true)
  * Dial in PIN `1401 7587 98` for [ETCCoreDevsCall13](https://meet.jit.si/static/dialInInfo.html?room=ETCCoreDevsCall13) numbers
* Focus: Mining Algorithm, Last-Call Proposals

### Agenda
* ECIP 1043 Fixed Dag Size Restriction (Last Call)
* ECIP 1099 Calibrate Epoch Duration (Draft, alternative to 1043)
[...]

PR #371, opened 2020-09-11
ECIP 1099: Add block numbers and move to Last Call

q9f:
ref #368 as per decision at call #362 (13)

moving 1099 to last call.

adding block number 11_700_000 for etc mainnet, 2_520_000 for mordor testnet.

cc @iquidus as ecip author

cc @meowsbits for core geth (needs activation on testnet for further evaluation)

cc @gregthegreek for hyperledger besu (same) and @bobsummerwill


iquidus (approved):
LGTM

YazzyYaz (approved):

bobsummerwill (changes requested):
Just playing catch-up here, it is not realistic that Besu code for ECIP-1099 will be ready on such a short timeline, and that will lead to Mordor being "broken" for Besu users.

I would like to suggest that we have an ephemeral testnet, so that Core-Geth developers can test their changes, and Besu can be added to that as soon as possible.   When things are looking healthy on that ephemeral testnet we can set an activation block on Mordor.

q9f:
Adding the `governance` label as you are raising concerns about a `last call` proposal.

If possible, please suggest more appropriate block numbers and we continue working on the 1099 implementation on a temporary testnet in parallel.

BelfordZ (approved):

q9f:
Putting this on hold for the next call to decide on a better timing.

q9f:
Updated the block numbers:

* For the Ethereum Classic main network: `ETCHASH_FORK_BLOCK := 11_700_000` (Epoch 390)
* For the Mordor Classic test network: `ETCHASH_FORK_BLOCK := 2_520_000` (Epoch 82)

Which basically means 4 weeks for Mordor and 8 weeks for the Classic main network. 

We will discuss this tomorrow in #362

BelfordZ:
LGTM 

blocked by @bobsummerwill - any update?

realcodywburns:
will the berlin HF eips be included in this fork or will there be another later fork for bls12-381 and simple subroutines? Or do we need a thread for the HF separate from this individual ECIP?

bobsummerwill:
@BelfordZ "blocked by @bobsummerwill - any update?"

Still waiting on ChainSafe, I am afraid.

CC @gregthegreek.

q9f:
No rush to merge this, but would be nice if we can agree on numbers later today.

@realcodywburns there is no Berlin hardfork specification yet, so I don't think this will happen this year.

bobsummerwill (approved):

Issue #374, opened 2020-09-12
[Discussion] ECBP-1100: Modified Exponential Subjective Scoring


meowsbits:
Discussion for https://github.com/ethereumclassic/ECIPs/pull/373, and an update to that original at https://github.com/ethereumclassic/ECIPs/pull/384.

realcodywburns:
It seems a little convoluted,  but it will eventually resolve to the 'correct' chain if given malicious reorg.

 > "Reasonable operating bounds for the victim of an eclipse attack to eventually recover and rejoin the network. " 

31 x difficulty ceiling seems overly cautious, given the bounds of the difficulty adjustment algorithm and could cause longer than desirable delays before a node would reorg to a proper chain, but it does allow for eventual resolution.  

Since this is an ecbp, the activation block could be genesis or any random block and, since this is only active on a synced chain, have the same effectiveness.  The coordination across nodes isnt as important since this allows a single node a ojective evaluation of incoming reorg blocks with eventual resolution even if it is 'wrong'. 

This solution could work on a network with some nodes using the 'pirlguard'  proposal and makes the same assumption that as long as a majority of nodes are evaluating reorgs with the assumption they are unusual outside of some window (`n` blocks for pg or 200 sec for mess).

realcodywburns:
maybe there is a need for a "core principle best practice" : "given a reorg of unusual length; nodes should evaluate the  chain head they have currently with preference over the incoming chain with longer length or higher difficulty"  This would cover PG, MESS, and possibly checkpointing.  

Your node is your castle, defend it from reorgs how you like. But if you have no defence (mess or pg not active) your node is less secure and possibly on the wrong chain.

As long as the majority of nodes are suspicious of reorgs, the network is more  resistant to malicious reorgs

meowsbits:
> Maybe there is a need for an overarching "core principle / best practice" : "small reorgs are normal and healthy, large reorgs are suspicious.

Yes!

> 31 x difficulty ceiling seems overly cautious

Understood. I wanted to start with the "most reasonably conservative" value and work our way toward efficiency from there. My  reasoning being that an eclipse of one or a few (arguably ill-equipped) nodes is inconvenient but recoverable, but a too-low ceiling could be permanently bad for everybody. Do you have a counteroffer? 22x? 11x? 3x? Keep in mind, too, that the value may be quite safely modified later if desired.

[...]

PR #384, opened 2020-09-22
ECIP/ECBP-1100 MESS: adopt polynomial function, update code

meowsbits:
- replaces _sin_ function with Vitalik's cubic antigravity function
- updates presented implementing code accordingly
- tries to better tackle the conceptual/principle part of it
- moves tentative mainnet "activation" block number to `11377500 (ETA Wednesay, 29 September 2020)` (again, not a hard fork, but configured in the same way)
- generally more and hopefully more useful implementation/specification details (eg for Besu devs)

:page_facing_up: View it [rendered](https://github.com/meowsbits/ECIPs/blob/ecbp/mess-2/_specs/ecip-1100.md).

### DIFF
--- a/_specs/ecip-1100.md
+++ b/_specs/ecip-1100.md
@@ -175,7 +175,7 @@ def get_curve_function_numerator(time_delta: int) -> int:
 __Condition `ecbp1100`__
 
 ```python
-if proposed_subchain_td * CURVE_FUNCTION_DENOMINATOR < get_curve_function_numerator(proposed.Time - commonAncestor.Time) * local_subchain_td:
+if proposed_subchain_td * CURVE_FUNCTION_DENOMINATOR < get_curve_function_numerator(current.Time - commonAncestor.Time) * local_subchain_td:
     return true
 return false
 ```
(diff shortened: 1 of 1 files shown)


developerkevin:
Isn't this moving a little too quickly for a mainnet activation on Sep. 29? If the network is attacked only core-geth nodes running MESS will remain on a chain while the other clients fracture off. Could this be something we discuss as a affected stakeholders and node operators? After all this did not go through the various stages of attaining a rough consensus per each graduation of ECIP levels as it is not exactly a hard fork.as stated above.

We still need assurances from those running core-geth clients that they've updated. We at the Co-op need to actually implement this into Besu if we are to reduce the risk of network fracturing given a 51% attack. I'm looking forward to discussing this tomorrow at 7AM PST.

ondertitels:
Frankly, I don't think this will cause any problems, as a large part of the nodes are run by exchanges. Not to mention that most exchanges are eagerly awaiting a fix to make the network more secure.

realcodywburns:
> 454 coregeth nodes

these are not all on the current version of geth. 

But even if it is only adopted by a minority of nodes, they will eventually follow to the canonical head using mess. Once the majority subjectively accept blocks, it will protect the chain.  

given the gravity of the vulnerability this is patching it seems like an acceptable risk.

realcodywburns (approved):
lgtm

meowsbits:
# Discussions to: https://github.com/ethereumclassic/ECIPs/issues/374

TokenHash:
<small>ok.</small>

q9f (approved):

PR #400, opened 2020-12-03
Admin Clean Up on ecip-1049.md

gitr0n1n:
+ move to Draft Status (Expired Last Call Review on Oct 9, 2020) due to material redraft of proposal.
+ Update Discussion Thread to new Redraft thread
+ Add Related Discussions Section for archived threads

### DIFF
--- a/_specs/ecip-1049.md
+++ b/_specs/ecip-1049.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1049
 title: Change the ETC Proof of Work Algorithm to Keccak256
-status: Draft
+status: Rejected
 type: Standards Track
 category: Core
 discussions-to: https://github.com/ethereumclassic/ECIPs/issues/394
--- a/_specs/ecip-1049.md
+++ b/_specs/ecip-1049.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1049
 title: Change the ETC Proof of Work Algorithm to Keccak256
-status: Rejected
+status: Draft
 type: Standards Track
 category: Core
 discussions-to: https://github.com/ethereumclassic/ECIPs/issues/394
@@ -11,7 +11,7 @@ created: 2019-01-08
 ---
 
 ### Abstract
-A proposal to replace the current Ethereum Classic proof of work algorithm with Keccak-256.
+A proposal to replace the current Ethereum Classic proof of work algorithm Etchash with Keccak-256.
 
 ### Motivation
 * A response to the recent double-spend attacks against Ethereum Classic. Most of this hashpower was rented or came from other chains, specfically Ethereum (ETH). A seperate proof of work algorithm would encourage the development of a specialized Ethereum Classic mining community, and blunt the ability for attackers to purchase mercenary hash power on the open-market. 
@@ -81,4 +81,5 @@ The area where most of the changes will be in `multi-geth/consensus`
 5. https://github.com/ethereumclassic/ECIPs/issues/362
 6. https://github.com/ethereumclassic/ECIPs/issues/382
 7. https://github.com/ethereum/EIPs/issues/2951
-8. https://github.com/ethereumclassic/ECIPs/issues/394
+8. https://vimeo.com/464336957
+9. https://github.com/ethereumclassic/ECIPs/issues/394
--- a/_specs/ecip-1049.md
+++ b/_specs/ecip-1049.md
@@ -2,11 +2,10 @@
 lang: en
 ecip: 1049
 title: Change the ETC Proof of Work Algorithm to Keccak256
-status: Last Call
-review-period-end: 2020-10-09
+status: Draft
 type: Standards Track
 category: Core
-discussions-to: https://github.com/ethereumclassic/ECIPs/issues/13
+discussions-to: https://github.com/ethereumclassic/ECIPs/issues/394
 author: Alexander Tsankov <alexander.tsankov@colorado.edu>
 created: 2019-01-08
 ---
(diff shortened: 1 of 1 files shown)


q9f:
I think we can continue working on that document, so I would rather send it back to draft status rather than rejecting it.

gitr0n1n:
> I think we can continue working on that document, so I would rather send it back to draft status rather than rejecting it.

Can that direct movement to `Draft` be made? source: the ECIP diagram.

I see that `Last Call` status can only be moved to `Accepted` or `Rejected`.

`Last Call - An ECIP that is done with its initial iteration and ready for review by a wider audience.`
https://github.com/ethereumclassic/ECIPs/raw/master/assets/ecip-1000/process.png

Pedantic, but thats how the Editors have chosen to follow the ECIP process for all ECIPs.

So move ECIP-1049 to Rejected. Alex resubmits the materially changed `reDraft` proposal to Draft status. Will that require a new ECIP # (like all the Aztlan proposals)?  @q9f

kizoojun01:
kizoojun@kizoojunkoi.org

kizoojun01 (changes requested):
kizoojun@kizoojunki.org

gitr0n1n:
> I think we can continue working on that document, so I would rather send it back to draft status rather than rejecting it.

@q9f can we get this ECIP moved back to the status it is currently in: `Draft`.

I believe we might needed to have it `Withdrawn` then resubmitted as `Draft`. The issue is that this proposal has past its current review period and now the authors are re-drafting the proposal. It's almost been in this limbo state for 60 days. I think its time to clean this up. I'd do it myself if I was an editor, but those PRs are also sittingon the shelf.

Thanks in advance for the ECIP housekeeping.

q9f:
I would supporting moving it back to Draft for the time being

bobsummerwill:
That would make sense to me too, @q9f.

The flow chart in ECIP-1000 does not show a path from Last Call back to Draft, but I don't think we gain anything by forcing the creation of a new, identical ECIP just for the sake of rigid adherence.

TokenHash:
@gitr0n1n @q9f @bobsummerwill ECIPs can perfectly be moved back to draft. The ECIP process does not specifically say that it can't and the drawing is just a diagram. I drew it like that cause I was copying the Bitcoin one (see below) and for some reason put arrows instead of just lines. I guess I unconsciously put arrows as I was thinking in the forward process, and not about special cases where ECIPs needed to go backward.

bip-process

The ECIP process has predecessors in the BIP and EIP processes. EIP says clearly:

```Last Call - This is the final review window for an EIP before moving to FINAL. An EIP editor will assign Last Call status and set a review end date (review-period-end), typically 14 days later.

If this period results in necessary normative changes it will revert the EIP to REVIEW.

Final - This EIP represents the final standard. A Final EIP exists in a state of finality and should only be updated to correct errata and add non-normative clarifications.

Stagnant - Any EIP in DRAFT or REVIEW if inactive for a period of 6 months or greater is moved to STAGNANT. An EIP may be resurrected from this state by Authors or EIP Editors through moving it back to DRAFT.```

BIP say clearly:

```BIPs should be changed from Draft or Proposed status, to Rejected status, upon request by any person, if they have not made progress in three years. Such a BIP may be changed to Draft status if the champion provides revisions that meaningfully address public criticism of the proposal, or to Proposed status if it meets the criteria required as described in the previous paragraph.```

I tried to change the diagram when Wei Tang was harassing the ETC ecosystem arguing the same thing about no backward status changes. The supposed impossibility of moving ECIPs backward to draft was always false, because it is obvious that when an ECIP has a catastrophic problem, no matter what is the status, they have to be moved back to draft to correct issues (as prescribed by the BIP and EIP). 

This has happened in ETH, but I don't remember what case it was.

In conclusion, ECIPs can be moved to draft (I co-wrote ECIP-1000). I will enter the corrected diagram to the assets folder so it is corrected and does not create these confusions again.

Wei was controlling the ECIP-1000 repo at the time of the argument and constantly sabotaging other ECIPs and commentary as he had editor, writing, and author access and authority. That is why I could never change the diagram.

**Comment by TokenHash** [comment]
@gitr0n1n @q9f @bobsummerwill  

This is the correct diagram for ECIP-1000:

process

...but I think I messed up when creating the PR from my fork (I haven't used GitHub for a long time  ;)

could one of you add the process.png image to the assets folder pls?

File: _specs/ecip-1049.md [hunk 1]
```diff
@@ -2,11 +2,10 @@
 lang: en
 ecip: 1049
 title: Change the ETC Proof of Work Algorithm to Keccak256
-status: Last Call
-review-period-end: 2020-10-09
+status: Rejected
```
TokenHash, line 5: Please change this to `Draft`.

gitr0n1n: Resolved.

gitr0n1n:
> @gitr0n1n @q9f @bobsummerwill
> 
> This is the correct diagram for ECIP-1000:
> 
> process
> 
> ...but I think I messed up when creating the PR from my fork (I haven't used GitHub for a long time ;)
> 
> could one of you add the process.png image to the assets folder pls?
> 
> Note that the diagram above is the new correct one, the previous one was not even updated to all the possible statuses that ECIPs may have.

Thanks for clarifying that diagram.

TokenHash:
@gitr0n1n I just re-learned how to use GitHub....lol....so I entered a PR to update the diagram.

TokenHash (approved):

bobsummerwill:
Wei must have polluted my brain with his rigidity about the process, @TokenHash!

q9f (approved):

q9f:
Thanks!

Miami87 (approved):

PR #406, opened 2021-02-03
Adding corrected process diagram

TokenHash:
Adding process-png diagram.


TokenHash:
I think this is including dome old PR, but the only part that matters is the change of the process-png on the assets folder of ecip-1000. Sorry.

TokenHash:
Closing because this one is no good. Included an old PR which should not be in this one.

PR #408, opened 2021-02-03
ECIP-1102: Use The Etherplan Consensus Engine for Approving Proto Treasury Funding Proposals

TokenHash:
Proposing The Etherplan Consensus Engine so the ETC community may use it to distribute proto treasury funds to projects that benefit ETC.


TokenHash:
Replacing this ECIP with a new one.

PR #411, opened 2021-02-09
ECIP 1103: Magneto EVM and Protocol Upgrades

q9f:


TokenHash (approved):

TokenHash:
lgtm

gitr0n1n (approved):
LGTM.

gitr0n1n:
@q9f Does Mantis support Berlin upgrades as well? Not sure if they've been pinged. @bpmckenna

bobsummerwill:
@gitr0n1n There is not Berlin support in Mantis, no.   Unlike the other clients, Mantis is ETC specific, so there is no upstream to pull from.

Also worth knowing is that Berlin is not even activated on ETH Mainnet yet.

gitr0n1n:
> @gitr0n1n There is not Berlin support in Mantis, no. Unlike the other clients, Mantis is ETC specific, so there is no upstream to pull from.
> 
> Also worth knowing is that Berlin is not even activated on ETH Mainnet yet.

Let me rephrase. This PR state Besu and Core-Geth intend to support Berlin upgrades. Does Mantis also intend to support Berlin EIPs?

developerkevin (approved):
LGTM.

PR #450, opened 2021-11-18
Withdraw ECIP-1084 "Hard-cap gas-limit in protocol".

bobsummerwill:
There was no consensus around this change, which is nearly two years old now.


stevanlohja (approved):
Approved. ECIP author has the right to withdraw their own ECIP. Thanks for the update!

PR #452, opened 2021-12-08
[ECIP-1104] BASEFEE omit, activations, last call

diega:
This PR proposes the following changes:
  - Exclude EIP-3198: `BASEFEE` Opcode from Mystique 
  - Add block number activation for each network based on the following calculations
    * **Mordor** (1587460 seconds until Jan 10th 2022): `5_294_406 block + 1_587_460 secs / 13.3 secs/block ~= 5_414_000 block`
    *  **Kotti** (2449694 seconds until Jan 20th 2022): `5_395_501 block + 2_449_694 secs / 15 secs/block ~= 5_559_000 block`
    *  **Mainnet** (4264005 seconds until Feb 10th 2022): `14_181_120 block + 4_264_005 secs / 13.3 secs/block ~= 14_502_000 block`
  - Change ECIP status to _Last call_

Ref: discussion #440

PR #459, opened 2022-01-17
Move ECIP-1104 to status Accepted

diega:
I propose to _accept_ the current state of the ECIP

### DIFF
--- a/_specs/ecip-1104.md
+++ b/_specs/ecip-1104.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1104
 title:  Mystique EVM and Protocol Upgrades
-status: Last Call
+status: Accepted
 type: Meta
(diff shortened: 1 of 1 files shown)


bobsummerwill:
I do not have permissions to review, but I agree with the change.

meowsbits (approved):
LGTM. I believe `Accepted` can only be applied 2 weeks after the move to `Last Call`.

diega:
Excellent. `Last Call` was set 11 days ago in #452. I think we can wait 3 more days

bobsummerwill:
That makes sense to me, yes.

bobsummerwill:
Actually, I see no real benefit in waiting just to hit that arbitrary 14 day point, so I will merge now.
That will help with the adoption.

PR #464, opened 2022-02-16
Add Bob Summerwill as author for ECIP-1049

antsankov:
Bob Summerwill is now currently championing ECIP-1049 and is an author of the proposal.


bobsummerwill (approved):
LGTM.

meowsbits (approved):

PR #468, opened 2022-02-22
removing myself from ecip editors

soc1c:
please update permissions accordingly

### DIFF
--- a/_specs/ECIP-0001.md
+++ b/_specs/ECIP-0001.md
@@ -81,7 +81,6 @@ The current ECIP editors are:
 
 * Mr. Meows D. Bits (@meowsbits)
 * Cody Burns (@realcodywburns)
-* Talha Cross (@soc1c)
 * Yaz Khoury (@YazzyYaz)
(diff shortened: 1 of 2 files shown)


gitr0n1n (changes requested):
@soc1c this edited document is ecip-0001.

Can you edit the active ecip-1000 document and move your name from the "Past Editors" list form the "Current Editors"

gitr0n1n:
Closing this PR in favor of #469 to edit the proper document. active ECIP-1000 (Wei) vs the withdrawn ECIP-0001 (Donald)

gitr0n1n:
Confirmed, permissions are update according to: 
https://github.com/orgs/ethereumclassic/teams

gitr0n1n (approved):
LGTM- updated the accurate document with the requested changes. It is confusing how Donald named the ECIP-0001 out of order. Seen this confusion a few times in commits. Perhaps we can rename that ECIP-0001 document to a different number that is more inline with all the other ECIPs since it was drafted/withdrawn later than ECIP-1000.

PR #470, opened 2022-02-25
Former Editors Alphabetize

gitr0n1n:
Former Editors Alphabetize by Handle

Clean up an old PR from Stev. Alphabetize the Former Editor list like the Current Editor list sits.

### DIFF
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -72,20 +72,20 @@ A good reason to transfer ownership is because the original author no longer has
 
 Current ECIP editors:
 
-* Antoine Toulme (@atoulme)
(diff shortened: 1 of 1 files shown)


bobsummerwill:
I don't know that this PR adds much value.

It took me a minute or two to even work out what the alphabet ordering even was.
Not first name, not surname, oh, it's based on github handle.

gitr0n1n:
> I don't know that this PR adds much value.
> 
> It took me a minute or two to even work out what the alphabet ordering even was. Not first name, not surname, oh, it's based on github handle.

Good point, likely an order issue. Fixed. 👍

IstoraMandiri (approved):

PR #471, opened 2022-02-25
Add Team of Current Editors

gitr0n1n:
Add Team of Current Editors


gitr0n1n:
It was noted that Antoine Toulme (@atoulme) does not have ECIP Editor perms to date. From @realcodywburns screenshot these perms may be pending acceptance.

bobsummerwill:
Antoine is not a member of ETC Cooperative staff.

He has executed on two short-term contracts paid by the Coop, firstly related to general mining support improvement for Besu and secondly to add Keccak support to Besu and Core-Geth.

IstoraMandiri (approved):

PR #477, opened 2022-02-26
Fix 1107 order appearance on web

gitr0n1n:
ECIP1107.md to ecip1108.md

Thought is that the uppercase filename is making this ECIP populate out of order in the lists on ecips.ethereumclassic.org. With this IM change it should fix that issue.

### DIFF
--- a/_specs/ecip-1107-rename.md
+++ b/_specs/ecip-1107-rename.md
@@ -1,7 +1,7 @@
 ---
 lang: en
 ecip: 1107
-title: ECIP Process
+title: ECIP Process (2020 Update)
 status: Withdrawn
 type: Meta
 discussions-to: https://github.com/ethereumclassic/ECIPs/issues/276
--- a/_specs/ecip-1108.md
+++ b/_specs/ecip-1108.md
@@ -1,7 +1,7 @@
 ---
 lang: en
-ecip: 1107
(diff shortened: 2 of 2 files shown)


atoulme (changes requested):
This updates the number of the ECIP, that seems wrong.

IstoraMandiri:
Right; why change the number and not just lowercase the filename?

gitr0n1n:
@IstoraMandiri @atoulme does it make sense that it is sorting to the top due to the `ECIP` being uppercase vs all others being lowercase? I'm trying to trouble shoot what i see on the ecips.ethereumclassic.org

image


We just changed this from ECIP-0001.md to ECIP-1107.md
So I was trying ecip-1108.md to fix the case. Let me know if there is a different way that will trigger github to update the file name to be all lowercase.

IstoraMandiri:
Yeah, it probably is due to it being uppercase. 

The best way to see what works is to build it locally outside of github, but it might be annoying to get an environment set up for for non developers. I'll investigate a bit further see if it's worth adding a docker container to make local development much easier.

You can trigger rebuilds on github in the Actions tab https://github.com/ethereumclassic/ECIPs/actions

gitr0n1n:
Fixed it!

PR #479, opened 2022-02-28
Fix author links

gitr0n1n:
Combed the http://ecips.ethereumclassic.org/all for broken author links or mismatched links and updated.

### DIFF
--- a/_specs/ecip-1010.md
+++ b/_specs/ecip-1010.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1010
 title: Delay Difficulty Bomb Explosion
-author: Igor Artamonov <splix@ethereumclassic.org>
+author: Igor Artamonov (@splix)
 status: Final
 type: Standards Track
 category: Core
--- a/_specs/ecip-1013.md
+++ b/_specs/ecip-1013.md
@@ -5,7 +5,7 @@ title: ETC On-Chain Cryptographic Signing and Authentication Protocol
 status: Draft
 type: Standards Track
 category: ECBP
-author: Cody W Burns <cody.w.burns@ethereumclassic.org>
+author: Cody Burns (@dontpanicburns)
 created: 2016-10-10
 ---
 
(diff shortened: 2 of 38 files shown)


IstoraMandiri:
It is not obvious that this is referring to GitHub user names, so perhaps we can link them automatically in future, but that would require a pull request to modify the build.

In the meantime, this seems like a reasonable approach.

IstoraMandiri (approved):

IstoraMandiri (changes requested):
I'm not sure about removing the emails, perhaps ask these authors first?

gitr0n1n:
> I'm not sure about removing the emails, perhaps ask these authors first?
> 
> I think a proper approach would be to include additional fields for Github ID and Emai, and show both in the UI somehow.

Yeah that would be a grat solution. Many of these emails are likely dead likely dead since they are on disbanded teams emails. I thought it was appropriate to update the emails to that authors most recent contact information.

IstoraMandiri:
Right, having checked it out, this is configured in https://github.com/ethereumclassic/ECIPs/blob/master/_includes/authorlist.html, which does support both right now depending on whether the name has an `(@` or an `<`, but will default to email if there's both.

While it is possible, I would rather not have to add additional UI elements as it clutters things up, so I can see why changing everything to github ids makes more sense.

I would like to hear some other maintainers thoughts about this?

meowsbits:
I don't like the change.

The documents in question were written and left as-is by the authors, including this field, their signature.
If you want to maintain an address book, start a new repo.

Further, modifying emails to Github-handles is backwards. The documents should want to be independent of the VCS web app we happen to be using at this point.

meowsbits (changes requested):
I don't like the change.

The documents in question were written and left as-is by the authors, including this field, their signature.
If you want to maintain an address book, start a new repo.

Further, modifying emails to Github-handles is backwards. The documents should want to be independent of the VCS web app we happen to be using at this point.

gitr0n1n:
Sounds good. I'll close this clean-up PR.

PR #480, opened 2022-02-28
Add github handles for authors

gitr0n1n:
Add github handles for authors


gitr0n1n:
We already do this in practice, so just updating the document to reflect the processes that are being followed to date as people choose their github name or their email.

IstoraMandiri (approved):

PR #482, opened 2022-02-28
Fix Formatting Errors

gitr0n1n:
Fix Formatting Errors on author fields

### DIFF
--- a/_specs/ecip-1077.md
+++ b/_specs/ecip-1077.md
@@ -1,7 +1,7 @@
 ---
 ecip: 1077
 title: The Meta Fork ECIP: Establishing _Fork_ type ECIPs and their specifications and processes
-author: Mr. Meows D. Bits (@meowsbits)
+author: Mr. Meows D. Bits
 discussions-to: TBD
 status: Withdrawn
 type: Meta
--- a/_specs/ecip-1077.md
+++ b/_specs/ecip-1077.md
@@ -1,7 +1,7 @@
 ---
 ecip: 1077
 title: The Meta Fork ECIP: Establishing _Fork_ type ECIPs and their specifications and processes
-author: Mr. Meows D. Bits
+author: Mr. Meows D. Bits (@meowsbits)
 discussions-to: TBD
 status: Withdrawn
 type: Meta
--- a/_specs/ecip-1082.md
+++ b/_specs/ecip-1082.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1082
 title: Fork identifier (`forkid`) protocol fork list specification
-author: github.com/meowsbits
+author: Mr. Meows D. Bits (@meowsbits)
 status: Superseded
 superseded-by: 1091
 type: Standards Track
@@ -156,4 +156,3 @@ The Ethereum Classic disinflationary monetary policy is not considered a series 
 ### Implementation
 
 Respective reference implementations for `forkid` generation are documented in EIP-2124 and EIP-2364.
-
--- a/_specs/ecip-1083.md
+++ b/_specs/ecip-1083.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1083
 title: Remove Contract Size Limit
-author: Zachary Belford <@BelfordZ>, Mudit Gupta <@maxsam4>
+author: Zachary Belford (@BelfordZ), Mudit Gupta (@maxsam4)
 discussions-to: https://github.com/ethereum/EIPs/issues/1662
 status: Draft
 type: Standards Track
--- a/_specs/ecip-1085.md
+++ b/_specs/ecip-1085.md
@@ -2,7 +2,7 @@
 lang: en
 ecip: 1085
 title: Simple Subroutines for the EVM
-author: Greg Colvin (@gcolvin) and Bob Summerwill (@bobsummerwill)
+author: Greg Colvin (@gcolvin), Bob Summerwill (@bobsummerwill)
 status: Draft
 type: Standards Track
 category: Core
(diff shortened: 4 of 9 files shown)


bobsummerwill:
/raw PONG _GHMoaCXLT is how the author chose to identify for two of these ECIPs, and there is no need to DOX the author with another name.   Also, the name used in this PR is not a full name, either.

Best NOT to change this.

_specs/ecip-1077.md
```diff
@@ -1,7 +1,7 @@
 ---
 ecip: 1077
 title: The Meta Fork ECIP: Establishing _Fork_ type ECIPs and their specifications and processes
-author: Mr. Meows D. Bits
+author: Mr. Meows D. Bits (@meowsbits)
```
meowsbits, line 4:
```suggestion
author: Mr. Meows D. Bits
```

meowsbits (changes requested):
Why are parenthetical identities better than bracketed? 

EDIT
Oh, I see, you just invented this... https://github.com/ethereumclassic/ECIPs/blame/master/_specs/ecip-1000.md#L158

Maybe we could consider some kind of syntax for specifying the app-handle relationship, like

```
John Doe <johndoe123@hotmail.com>
Jonh Doe (github:johndoe)
```

The problem is that `@johndoe` is ambiguous/useless for the repo as a decentralizable collection of documents, since it presumes the existence of and integration with the Github corporation.

gitr0n1n:
> /raw PONG _GHMoaCXLT is how the author chose to identify for two of these ECIPs, and there is no need to DOX the author with another name. Also, the name used in this PR is not a full name, either.
> 
> Best NOT to change this.

This comment doesn't make much sense:
https://github.com/q9f
http://ecips.ethereumclassic.org/ECIPs/ecip-1103

_specs/ecip-1077.md
gitr0n1n: Just following the ECIP-1000

"The Author header lists the names and email addresses/GitHub handles of all the authors/owners of the ECIP. The format of the Author header value must be:

Random J. User <123@dom.ain>
Random J. User (@random.j.user)
If there are multiple authors, each should be on a separate line following RFC 2822 continuation line conventions."

gitr0n1n: Just following the ECIP-1000

The Author header lists the names and email addresses/GitHub handles of all the authors/owners of the ECIP. The format of the Author header value must be:

Random J. User <123@dom.ain>
Random J. User (@random.j.user)
If there are multiple authors, each should be on a separate line following RFC 2822 continuation line conventions.

gitr0n1n:
> Why are parenthetical identities better than bracketed?
> 
> EDIT Oh, I see, you just invented this... https://github.com/ethereumclassic/ECIPs/blame/master/_specs/ecip-1000.md#L158
> 
> Maybe we could consider some kind of syntax for specifying the app-handle relationship, like
> 
> ```
> John Doe <johndoe123@hotmail.com>
> Jonh Doe (github:johndoe)
> ```
> 
> The problem is that `@johndoe` is ambiguous/useless for the repo as a decentralizable collection of documents, since it presumes the existence of and integration with the Github corporation.

Just fixing broken links. Absent links. 👍

meowsbits:
Generally, editing other people signatures is going to feel wrong to me in 99% of the cases. We have no hard requirements for identitiy/entity assertion, nor should we.

gitr0n1n:
> Generally, editing other people signatures is going to feel wrong to me in 99% of the cases. We have no hard requirements for identitiy/entity assertion, nor should we.

These are not "signatures" these are ECIP `author` headers in the Preamble.

_specs/ecip-1077.md
meowsbits: I have a different interpretation: It does not mean than an author MUST register an email (OR a Github handle, now); it means that IF an author registers one of those values, they should be formatted as described.

gitr0n1n:
> Generally, editing other people signatures is going to feel wrong to me in 99% of the cases. We have no hard requirements for identitiy/entity assertion, nor should we.
> 
> > Just fixing broken links.
> 
> Where are they broken? Can you please describe the issue better?

Sure. Here is  an example of a broken link.

image

meowsbits:
That isn't a broken link because it isn't a link. Its plain text.

gitr0n1n:
> That isn't a broken link because it isn't a link. Its plain text.

Here is another for you.
image

meowsbits:
I am OK with punctuation-only edits. But changing, adding, or removing people's names is substantive, and not OK with me.

gitr0n1n:
> I am OK with metadata-syntax/punctuation-only edits. But changing, adding, or removing people's names (real or claimed) is substantive, and not OK with me.

Great we will start with that as I comb through this repo for generally immaterial admin clean up.

IstoraMandiri:
> That isn't a broken link because it isn't a link. Its plain text.

In this case it looks like the author wasn't was accidentally using the wrong syntax `(` instead of `<`, which is why the link is broken, see https://github.com/ethereumclassic/ECIPs/pull/479#issuecomment-1053909982

gitr0n1n:
> I agree, would you support overriding current `(@blah)` with this new `(github:blah)` or `(twitter:blah)`, and while we are at it should we replace `<hi@joe.com>` with `(email:hi@joe.com)` ?
> 
> This is clearly a confusing feature currently

I like this formatting you are suggesting because it is clearly readable in markdown text.
e.g.:  `(github:blah)` or `(twitter:blah)

meowsbits:
I don't opinions on the precise handle/tag syntax... Probably there are prior arts and patterns that we can reuse instead of reinventing... 

Also. just by the way, `ecip_validator` is a gem that codifies these kinds of validations, and is currently run in the Github Actions workflow. That may be a good place for some of these kinds of validations we're talking about here.

gitr0n1n:
So what's going on with this PR @ethereumclassic/ecip-editors ? We good to push and merge?

meowsbits (approved):

PR #497, opened 2022-10-26
Updates to the "Ethereum Classic Network Description" which is the canonical description of ETC hard forks

bobsummerwill:
Updates to the "Ethereum Classic Network Description" which is the canonical description of ETC hard forks.

- Added missing entry for EIP-160 (Spurious Dragon for ETC)
- Minor fix to URL for EIP-684.

We could really use proper "Meta" ECIPs to record these early hardforks, but that is a task for another day.

- "Gas Reprice" (ETC's Tangerine Whistle)
- "Die Hard" (ETC's Spurious Dragon and difficulty bomb delay)
- "Gotham" (Fixed monetary policy)
- "Defuse Difficulty Bomb"

### DIFF
--- a/_specs/ecip-1066.md
+++ b/_specs/ecip-1066.md
@@ -15,10 +15,11 @@ Ethereum Classic network mainnet.
 | ---------- | ------------ | ---------------------- | -------------------------------------------------------------- |
 | 2015-07-30 |            0 | Frontier               | [EIP-684](https://github.com/ethereum/EIPs/issues/684)         |
 | 2015-07-30 |            1 | 5M20 Era 1             | [ECIP-1017](https://ecips.ethereumclassic.org/ECIPs/ecip-1017) |
-| 2015-09-08 |      200,000 | Ice Age                | [EIP-684](https://github.com/ethereum/EIPs/issues/684)         |
+| 2015-09-08 |      200,000 | Ice Age                | [EIP-684](https://eips.ethereum.org/EIPS/684)         |
 | 2016-03-15 |    1,150,000 | Homestead              | [EIP-606](https://eips.ethereum.org/EIPS/eip-606)              |
 | 2016-10-24 |    2,500,000 | Gas Reprice            | [ECIP-1015](https://ecips.ethereumclassic.org/ECIPs/ecip-1015) |
 | 2017-01-13 |    3,000,000 | Die Hard               | [ECIP-1010](https://ecips.ethereumclassic.org/ECIPs/ecip-1010) |
+| 2017-01-13 |    3,000,000 | Spurious Dragon        | [EIP-160](https://eips.ethereum.org/EIPS/eip-160) |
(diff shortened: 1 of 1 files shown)


bobsummerwill:
@atoulme @gitr0n1n @IstoraMandiri @meowsbits

_specs/ecip-1066.md
```diff
@@ -15,10 +15,11 @@ Ethereum Classic network mainnet.
 | ---------- | ------------ | ---------------------- | -------------------------------------------------------------- |
 | 2015-07-30 |            0 | Frontier               | [EIP-684](https://github.com/ethereum/EIPs/issues/684)         |
 | 2015-07-30 |            1 | 5M20 Era 1             | [ECIP-1017](https://ecips.ethereumclassic.org/ECIPs/ecip-1017) |
-| 2015-09-08 |      200,000 | Ice Age                | [EIP-684](https://github.com/ethereum/EIPs/issues/684)         |
+| 2015-09-08 |      200,000 | Ice Age                | [EIP-684](https://eips.ethereum.org/EIPS/684)         |
 | 2016-03-15 |    1,150,000 | Homestead              | [EIP-606](https://eips.ethereum.org/EIPS/eip-606)              |
 | 2016-10-24 |    2,500,000 | Gas Reprice            | [ECIP-1015](https://ecips.ethereumclassic.org/ECIPs/ecip-1015) |
 | 2017-01-13 |    3,000,000 | Die Hard               | [ECIP-1010](https://ecips.ethereumclassic.org/ECIPs/ecip-1010) |
+| 2017-01-13 |    3,000,000 | Spurious Dragon        | [EIP-160](https://eips.ethereum.org/EIPS/eip-160) |
```
meowsbits, line 22: Maybe we can just add the EIP160 ref to the existing "Die Hard" fork? ETC only adopts 160, while the ETH namesake Spurious Dragon implements state trie clearing (EIP161, et al), which might be confusing since ETC implements EIP161 only with the Atlantis fork.

bobsummerwill: So maybe this table should just say "EXP cost increase" rather than "Spurious Dragon"?

Or better yet we create new (very simple) Metas for these forks and use them here, rather than the weird mix of plain ECIPs and Metas?

Because "Die Hard" in this table is not pointing to a spec for "Die Hard", but to one of two EIPs/ECIPs in that HF.

meowsbits: Basically I would shoot to make a table like the one here https://github.com/ethereum/execution-specs#readme

If there's no Meta ECIP (eg. HFM-xxx), then I'd just leave it blank for now to minimize scope of change. This table, if it gets filled in with the missing other half of the EIPs that got activated, does the job of the Meta ECIP nearly as well.

meowsbits: See https://github.com/ethereumclassic/ECIPs/pull/499 for the bespoke table.

bobsummerwill:
Withdrawn in favor of https://github.com/ethereumclassic/ECIPs/pull/499.

PR #499, opened 2022-10-28
ECIP-1066: revise table to show EIP/ECIP::Fork relationships

meowsbits:
Intended to offer a candidate to continue and replace #497. Makes the table the same format as the table found at https://github.com/ethereum/execution-specs/blob/333baf195cb4c3d3f7ab19698443108694a71d8f/README.md. 

View the rendered file [here](https://github.com/meowsbits/ECIPs/blob/network-fork-specs/_specs/ecip-1066.md).


meowsbits:
@bobsummerwill Note that I've left off the 5M20 table entries that would occur every 5 million blocks on the ECIP-1017 schedule. These are kind of an "upgrade" gray zone, aren't they? :wink: :grimacing: Left them out because... I don't want to have to come back here in another 6 years just to add another row to this table... :)

bobsummerwill:
Yes, this is way, way better.
Let's do this instead.
And I would agree that we don't need to include those 5M20 entries - they are implicit and are not protocol changes.

bobsummerwill (approved):
This is way better than the existing table.
LGTM - thanks!

bobsummerwill:
PS.  I there an entry for MESS here because we have tests for it?

Because it's not a protocol change per se.

meowsbits:
Yea, its not a hard fork, but I put it in there because I figured people might look for it, and because whether or not its a "protocol upgrade" depends on the context's definition of "protocol" and "upgrade" :).

_specs/ecip-1066.md
```diff
@@ -11,24 +11,21 @@ created: 2019-10-07
...
+| MESS                                         | 11380000  | 2020-10-10 | [ECBP-1100](https://ecips.ethereumclassic.org/ECIPs/ecip-1100)                                                                                                                                                                                                                                                                                                                                                                                                                   |                                                                |                                                   |  |
```
bobsummerwill, line 19: So yeah - let's just remove this line, please, unless we specifically have ethereum/tests for it?
It is not a part of the consensus protocol.
Also (hopefully) it will be gone soon enough anyway.

meowsbits, line 19:
```suggestion
```

meowsbits: Done with https://github.com/ethereumclassic/ECIPs/pull/499#pullrequestreview-1167038662

PR #504, opened 2023-03-14
Update ecip-1066.md

gitr0n1n:
Add authors.


gitr0n1n:
@ethereumclassic/ecip-editors it looks like the build is failing due to a deprecated element. Can one of you take a peek at it and see what is needed to get the automation updated? Much thanks!

Issue #505, opened 2023-03-18
Remove atoulme from ECIP editors


atoulme:
Hello all,

I have not been active and given my lack of contributions, I think it is time for me to step down as ECIP editor. Please let me know if this is appropriate and accept my resignation. I will follow steps per process.

TheCrowbill:
This issue should have closed with the merge of #508 but I am unable to link.

PR #508, opened 2023-04-12
atoulme ECIP Editor resignation

gitr0n1n:
Source: https://github.com/ethereumclassic/ECIPs/issues/505

- Move atoulme to Former editors list.
+ Add r0n1n's public ETC-centric organization.

@realcodywburns will you remove @atoulme from @ethereumclassic/ecip-editors access? Also while you're in there can you double check permissions in the repo and remove stale/inactive accounts that may have access due to involvement in the past? Much thanks!

### DIFF
--- a/_specs/ecip-1000.md
+++ b/_specs/ecip-1000.md
@@ -72,14 +72,13 @@ A good reason to transfer ownership is because the original author no longer has
 
 Current ECIP editors:
 
-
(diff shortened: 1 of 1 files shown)


atoulme (approved):

realcodywburns:
ack. need to updste ruby...

PR #520, opened 2023-12-11
ECBP-1110: Deactivate MESS

meowsbits:
Turn off the artificial finality mechanism [ECBP-1100's MESS (Modified Exponential Scoring)](https://ecips.ethereumclassic.org/ECIPs/ecip-1100) in [Core-Geth](https://github.com/etclabscore/core-geth) for Ethereum Classic.

Rendered document: https://github.com/meowsbits/ECIPs/blob/ecbp1110/_specs/ecip-1110.md

FYI I'll also be updating https://meowsbits.github.io/51-percent-docs/ this week, which provides some useful context.

### DIFF
--- a/_specs/ecip-1100.md
+++ b/_specs/ecip-1100.md
@@ -4,7 +4,7 @@ title: MESS (Modified Exponential Subjective Scoring)
 lang: en
 author: Isaac <b5c6@protonmail.com>
 discussions-to: https://github.com/ethereumclassic/ECIPs/issues/374
-status: Active
+status: Replaced
 type: Standards Track
 category: ECBP
 created: 2020-09-09
@@ -14,6 +14,8 @@ license: Apache-2
 ## Abstract
 
 Define a function arbitrating chain acceptance using relative total difficulty and common ancestor time to raise finality confidence.
+
+_Update (January 2024):_ Replaced with [ECBP-1110](https://ecips.ethereumclassic.org/ECIPS/ecip-1110).
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


gitr0n1n:
To clarify @meowsbits, MESS still exists in the client as a tool should it ever be needed again by client users, but this removes the default activation setting and adjusted the setting to "off" by default?

bobsummerwill:
Discussion ongoing here:  https://github.com/orgs/ethereumclassic/discussions/522

gitr0n1n:
Now that we have a a block set for Mordor. Is this ECIP in a final state for review? @meowsbits Ping me and I'll be happy to approve since there does not appear to be any contention with this ECIP proposal and discussion has been open at:
https://github.com/orgs/ethereumclassic/discussions/522

gitr0n1n (approved):
Mordor activation set 10 hours ago. Discussion on the topic seems to support this change without any contention. My understanding is the MESS tool is not being removed at the moment, but its default is OFF due to the abundance of friendly/loyal hashrate that is securing the ETC network since Sept 2022.

gitr0n1n:
> Also for status, do we want to update 1100 with "Replaced by 1099" or something to that effect? Or keep as active, since it is still present in the client, but just not turned on by default?

@meowsbits can you comment on this? How should we adjust the status?

meowsbits:
I like your `Replaced` idea on 1100, the original MESS (activation) spec, concurrent with this. Although I think leaving it would be logical too, I guess it will be more understandable if it's association with this 1110 spec is documented.  Will do.

meowsbits:
CI failed because Ruby's `bundler` install failed. Unrelated to changes here, so I'll go ahead with merge and fix the bundler issue later.

https://github.com/ethereumclassic/ECIPs/actions/runs/7463281525/job/20307577984?pr=520
```
Run gem install bundler
ERROR:  Error installing bundler:
	The last version of bundler (>= 0) to support your Ruby & RubyGems was 2.4.22. Try installing it with `gem install bundler -v 2.4.22`
	bundler requires Ruby version >= 3.0.0. The current ruby version is 2.6.10.210.
Error: Process completed with exit code 1.
```

PR #527, opened 2024-02-25
Update ECIP Editors to perms

chris-mercer:
Update list to accurately reflect the perms in the repo.


realcodywburns (approved):
I am happy to participate in an editor role again. Ack swapping ronin and chris

PR #547, opened 2025-12-05
Create ECIP-1120: Basefee Market with Miner Rewards

IstoraMandiri:
This WIP ECIP introduces a minimally complicated implementation of [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) for Ethereum Classic. 

ECIP-1120 gives users the option to price transactions using an algorithmic Basefee Market, via the opt-in "Type 2" transaction. 

It proposes that the Basefee is distributed in full to miners using a protocol-native smooth distribution mechanism.

A compendium research and information website is available at https://ecip1120.dev.

Note: The `1120` ECIP number was approved by @realcodywburns to give room for other ECIP authors using the `111X` namespace.

PR #550, opened 2025-12-14
Fix Olympia ECIPs categorization

IstoraMandiri:
This PR reclassifies Olympia related ECIPs correctly to follow their ECIP types and accurately reflect their WIP status.

- All Olympia ECIPs are `WIP` status, since they are missing key research tasks, specifically around Block Elasticity, as discussed in https://github.com/ethereumclassic/ECIPs/pull/547 
- 1111 and 1112 type is `Standards Track` / `Core`  as they require a hard fork
- 1113 and 1114 type remains `Meta` as they are not requiring a hard fork (contract layer)

### DIFF
--- a/.github/workflows/ecip_validation.yml
+++ b/.github/workflows/ecip_validation.yml
@@ -15,4 +15,4 @@ jobs:
       run: |
         gem install bundler -v 2.4.22
         bundle install --jobs 4 --retry 3
-        shopt -s globstar; bundle exec ecip_validator _specs/**/*.md
+        shopt -s globstar; bundle exec ruby bin/ecip_validator _specs/**/*.md
--- a/bin/ecip_validator
+++ b/bin/ecip_validator
@@ -1,0 +1,23 @@
+#!/usr/bin/env ruby
+# Wrapper script that patches ecip_validator to add WIP as a valid status
+# per ECIP-1000 which states: "Draft ECIPs which may be in a very early stage
+# may be entered as WIP ECIPs"
+
+require 'bundler/setup'
+require 'ecip_validator'
+
+# Monkey-patch to add WIP status per ECIP-1000
+module EcipValidator
+  class Validator
+    # Override the validates_inclusion_of for status
+    _validators.delete(:status)
+    validates_inclusion_of :status, in: [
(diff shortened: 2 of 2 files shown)


IstoraMandiri:
Okay, I understand the issue much better now. 

@chris-mercer's rationale for incorrectly classifying does make a bit more sense now, though we could have acheived the same thing with a `[WIP]` prefix rather than reclassifying incorrectly.

Anyway, here's a summary of a patch that for now allows WIP ECIPs.

## Fix ECIP Validator to Accept WIP Status

### Problem

The CI validation was failing with:

```
_specs/ecip-1111.md is NOT valid:     {:status=>["is not included in the list"]}
_specs/ecip-1112.md is NOT valid:     {:status=>["is not included in the list"]}
_specs/ecip-1113.md is NOT valid:     {:status=>["is not included in the list"]}
_specs/ecip-1114.md is NOT valid:     {:status=>["is not included in the list"]}
```

The `ecip_validator` gem does not include "WIP" in its [list of valid statuses](https://github.com/meowsbits/ecip_validator/blob/master/lib/ecip_validator/validator.rb#L10), despite ECIP-1000 explicitly allowing it:

> `Draft` ECIPs which may be in a very early stage may be entered as `WIP` ECIPs, which means they are a work in progress.
> — ECIP-1000, line 215

### Solution

Added a local wrapper script (`bin/ecip_validator`) that:
1. Loads the upstream `ecip_validator` gem
2. Patches the `Validator` class to include "WIP" in the valid status list
3. Runs the validation

Updated the CI workflow to use this local script instead of calling the gem directly.

### Why Not Update the Upstream Gem?

The `meowsbits/ecip_validator` repository appears to be inactive with no recent maintenance. @meowbits is a former ECIP editor and is no longer available to update the gem.

### Future Work

We should consider replacing the `meowsbits/ecip_validator` dependency entirely with a locally-maintained validator in this repository. This would:

- Give us full control over validation rules
- Ensure the validator stays in sync with ECIP-1000
- Remove dependency on an unmaintained external gem
- Allow easier updates when the ECIP process evolves

For now, this wrapper script provides a minimal fix that unblocks the Olympia ECIPs (1111-1114) while maintaining compatibility with the existing gem.

PR #551, opened 2025-12-14
Fix Olympia ECIPs categorization

IstoraMandiri:
This PR reclassifies Olympia related ECIPs correctly to follow their ECIP types and accurately reflect their status.

- 1111 and 1112 type is `Standards Track` / `Core`  as they require a hard fork
- 1113 and 1114 type is `Standards Track` / `ECBP` as they define Application-level standards and conventions


chris-mercer:
Unnecessary. if you're changing/standardizing how things are categorized and such, a script can be applied to the whole ECIP stack so standards are applied in alignment to the text in the ECIP-1000. I believe I mentioned this in another thread. The goal should be consistency across all ECIPs.

Suggesting this PR be closed with the understanding that the ECIP repo should have have a scripted cleanup where standards are blindly applied to all ECIPs in compliance with ECIP-1000 classification. This removes the burden from ECIP editors and prevents the violation of ECIP norms and practices we saw with the ECIP-1120 proposal and approval.

This cleanup would also including fixing the build to accommodate WIP and other status. in 2026 we shouldn't need to format/lint ECIP front matter manually anymore. There is no reason this repo should be working on such dated code when its a simple markdown library. Probably under a day revamp this to modern tooling to fix all of this with Claude at this point.

PR #557, opened 2026-01-26
ECIP (draft) - Quantum-resistance for Ethereum Classic: ML-DSA verification precompile

GravityLabLLC:
WIP

Discussion: https://github.com/orgs/ethereumclassic/discussions/556

PR: https://github.com/ethereumclassic/core-geth/pull/5

### DIFF
--- a/_specs/ecip-1122.md
+++ b/_specs/ecip-1122.md
@@ -1,0 +1,242 @@
+---
+ecip: 1122
+title: ML-DSA (FIPS 204) Signature Verification Precompile(s)
+lang: en
+translator:
+author: Daniel Tocila
+discussions-to: https://github.com/ethereumclassic/ECIPs/discussions
+status: WIP
+type: Standards Track
+category: Core
+created: 2026-01-25
+requires: 1103
+replaces:
+superseded-by:
+resolution:
+---
+
+## Abstract
+
+This ECIP introduces one or more native EVM precompiles for verifying post-quantum digital signatures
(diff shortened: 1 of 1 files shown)


GravityLabLLC:
> Highlighting this comment so it doesn't appear that this wasn't looked at. Again, i think upstream EVM is the best place for this. They have the resources and bandwidth to implement this in a serious way. This project is currently in a funding crisis with its deprecated client in maintenance mode, so its unlikely this will get any attention atm. If this is adopted upstream it will make its way into ETC at some point should we figure out how to self-fund this project.
> 
> https://github.com/orgs/ethereumclassic/discussions/556#discussioncomment-15686274

Thank you for the comment. I responded there. The tl;dr is that I think there are plenty of people available with which to do code review once the testing is complete and there is a successful test deployment with ample results to share. This would mean ETC can lead in the space. We like the idea of self-funding of projects and look forward to exploring that further with conversations about Elysium and how it relates to Olympia in this way.

chris-mercer:
@GravityLabLLC, a numbering correction: the ECIP number in your title is being formalized for the Olympia network-security ECIP (referenced throughout client implementations and updated specs), and the 112x range is already the Olympia suite, with ECIP-1121 as the meta spec. Per ECIP-1000, ECIP numbers are assigned by editors at formal submission, not self-assigned, so I'm retitling this to the un-numbered draft format (matching your #568, #569, #570) until it's formally submitted.

For post-quantum work, a dedicated range (1140s) makes sense as more QR proposals come in, your ML-DSA precompile would be the anchor as the first QR proposal submitted. We can assign the number at formal submission.

PR #580, opened 2026-08-26
spec(ecip-1100,1110): ECBP-1100 is Active not Replaced

chris-mercer:
ECBP-1100 is marked `Replaced` and ECBP-1110 is titled "Deactivate MESS". Both misstate where the network actually stands: MESS is still present in Core-Geth and any operator can enable it. ECBP-1110 recommended that client developers ship the flag off by default — it did not remove the mechanism, and nothing replaced ECBP-1100.

>Abstract
>Turn off the artificial finality mechanism [ECBP-1100’s MESS (Modified Exponential Scoring)](https://ecips.ethereumclassic.org/ECIPs/ecip-1100) in [Core-Geth](https://github.com/etclabscore/core-geth) for Ethereum Classic.

### ecip-1100

- `status: Replaced` → `status: Active`. Nothing replaced it, and the mechanism is still in the client.
- The January 2024 update note read "Replaced with ECBP-1110". It now states that ECBP-1110 recommends the disabled default, that MESS remains available to any operator who enables it, and that ECBP-1100 therefore remains active and optional.

### ecip-1110

- Retitled "Deactivate MESS (Modified Exponential Subjective Scoring)" → "Disable MESS (Modified Exponential Subjective Scoring) as Default", so the title states what the ECBP recommends rather than confusing it with removal.
- The abstract now names the 2023 hashrate surge as the reason for the recommendation, and states that the mechanism remains available to any operator who enables it.
- "Deactivated" → "disabled" in the Specification and Mordor sections, so the document uses one verb throughout for one thing. The Mordor line also gains "by default", matching the mainnet line above it.

### Override flag name

The Specification documented the override flag as `--override.ecbp1100-deactivate`. Core-Geth v1.12.17 — the release this ECBP names — spells it with a dot:

```go
// cmd/utils/flags.go, v1.12.17
OverrideECBP1100DeactivateFlag = &cli.Uint64Flag{
	Name:     "override.ecbp1100.deactivate",
	Usage:    "Manually specify the ECBP-1100 (MESS) deactivation block number, overriding the bundled setting",
	Category: flags.EthCategory,
}
```

The hyphenated form appears nowhere in that release, so an operator copying it out of this document gets an unknown-flag error. Corrected to `--override.ecbp1100.deactivate`.

`--ecbp1100=99999999999` further down is left as written — it is the activation flag's real name at that release.

### Non-substantive

`ecip-1100` also picks up markdown normalization: `__bold__` → `**bold**`, `+` list markers → `-`, trailing whitespace removed, and blank lines inserted before some lists and block quotes. No wording changes, and no change to rendered output — the affected lists and quotes already rendered correctly.

### DIFF
--- a/_specs/ecip-1100.md
+++ b/_specs/ecip-1100.md
@@ -15,7 +15,7 @@ license: Apache-2
 
 Define a function arbitrating chain acceptance using relative total difficulty and common ancestor time to raise finality confidence.
 
-_Update (January 2024):_ Replaced with [ECBP-1110](https://ecips.ethereumclassic.org/ECIPs/ecip-1110).
+_Update (January 2024):_ [ECBP-1110](https://ecips.ethereumclassic.org/ECIPs/ecip-1110) turned the bundled default off. The mechanism remains in the client and available to any operator who enables it.
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


IstoraMandiri:
I think the correct approach here is to update the ECIP process and add a `deprecated` status.

Probably the intent of using `replaced` originally because it there isn't really a suitable status for this.

Active is technically true but is also confusing.

PR #583, opened 2026-08-27
fix(deps): clear all 23 dependency advisories add Dependabot config

chris-mercer:
`bundle-audit` reports **23 advisories** against the committed `Gemfile.lock`. This clears all of them.

| | advisories |
|---|---|
| before | **23** — nokogiri 12, concurrent-ruby 3, activesupport 3, json 2, faraday 2, addressable 1 |
| after | **0** |

## Bundler — a prerequisite, not hygiene

`Gemfile.lock` pinned `BUNDLED WITH 2.1.4`, which `ruby/setup-ruby` installs on every CI run (both workflows set `bundler-cache: true`). 2.1.4 carries **CVE-2021-43809** (argument-injection RCE via a dash-leading git URL in a Gemfile, fixed 2.2.33) and **CVE-2020-36327** (dependency confusion, fixed 2.2.10). Both workflows trigger on `pull_request`, so that Bundler runs against Gemfile content supplied by the pull request.

It is also load-bearing: with 2.1.4 in place, `bundle lock --update` fails outright with `Unable to find a spec satisfying nokogiri (>= 1.10.8) in the set`. Nothing below is reachable until Bundler moves.

## Lockfile only

The existing `Gemfile` constraints already permitted all of these:

| gem | from | to | advisories |
|---|---|---|---|
| nokogiri | 1.18.10 | 1.19.4 | 12, incl. HIGH CSS-selector ReDoS — parses on every `jekyll build` |
| activesupport | 8.1.1 | 8.1.3.1 | 3 (ReDoS, XSS, DoS) |
| concurrent-ruby | 1.3.6 | 1.3.8 | 3, incl. HIGH NaN livelock |
| json | 2.18.0 | 2.21.2 | 2, incl. HIGH format-string injection |
| addressable | 2.8.8 | 2.9.0 | 1 HIGH ReDoS in template expansion |

## Gemfile change — faraday

| gem | pinned | resolved | now | advisories |
|---|---|---|---|---|
| faraday | `~> 0.17.3` | **0.17.6** | `~> 1.10.6` | CVE-2026-25765 (SSRF), CVE-2026-54297 (HIGH, recursion DoS) |

The `0.17.3` pin dates from 2020 (`4b1e555`, *"try pinning faraday dep to fix jekyll build"*) and resolved to **0.17.6** — the version actually locked and audited.

`sawyer`, via `octokit`, requires `faraday (> 0.8, < 2.0)`. That range *includes* 0.17.6, so the constraint is not what rules the old line out. **1.10.6 is the lowest release clearing both defects in the maintained 1.x line** — the SSRF is fixed in 1.10.5, the DoS in 1.10.6.

### Why the two advisory databases disagree about 0.x

Worth stating, because the counts above and the Gemfile comment look contradictory otherwise:

| | CVE-2026-25765 | CVE-2026-54297 | CVE-2026-33637 |
|---|---|---|---|
| ruby-advisory-db (what `bundler-audit` reads) | `patched: ~> 1.10.5, >= 2.14.1`, no unaffected clause → **0.17.6 flagged** | `patched: ~> 1.10.6, >= 2.14.3` → **0.17.6 flagged** | `unaffected: < 2.0.0` → 0.17.6 clean |
| OSV / GHSA | `introduced: 1.0.0` → 0.x outside every range | `introduced: 1.0.0` → 0.x outside every range | `introduced: 2.0.0` only |

Both readings are defensible and neither is a reason to stay on 0.x. **Read 0.x as never backported rather than as unaffected:** it is end-of-life at 0.17.6 and carries both defects. It is the worse of the two on the SSRF — 1.10.6 rejects a protocol-relative String host but not a `URI` object (CVE-2026-33637, declared against 2.x only, so scanners report 1.10.6 clean), and 0.17.6 rejects neither.

The constraint is `~> 1.10.6` rather than `~> 1.10`, which would re-admit 1.10.0 through 1.10.5. The original pin's issue link is preserved in the Gemfile comment.

## Dependabot

`.github/dependabot.yml` is added so this does not silently age again — but **wired disabled, not active**.

`open-pull-requests-limit: 0` is GitHub's documented mechanism for keeping a configuration visible and reviewed while nothing fires. Nobody is triaging a standing pull-request queue against this repository yet, and a queue that is not read reports itself as a control while operating as noise. Enabling it later is a one-line change.

The cooldown values (7 days, 21 for majors) are set and inert until that limit is raised. Both keys govern *version* updates only — **Dependabot security updates are a repository setting and cannot be enabled from this file.**

## Verification

Run on Ruby 3.2.3 under the repo's frozen bundle config:

```
bundle install ..................... 129 gems, exit 0
bundle exec rake validate .......... 97/97 valid
bundle exec jekyll build --future .. exit 0
bundle-audit check ................. 23 advisories -> No vulnerabilities found
```

The built site is **identical across all 158 files** except `feed.xml`'s `<updated>` build timestamp. Liquid warning count is unchanged (2 before, 2 after — the pre-existing ecip-1048/1060 pair, addressed separately in #581).

Advisory data: `bundler-audit 0.9.3` against `rubysec/ruby-advisory-db` @ `eca0ecc` (1,234 advisories, updated 2026-08-23).

**Jekyll stays at 3.10.0 deliberately** — `github-pages 232` pins it to match what GitHub Pages itself runs, and no advisory affects Jekyll at that version.

**Not addressed here:** Ruby 3.2 reached EOL on 2026-04-01 and both workflows pin it. That is a separate change — none of the fixes above require it, since every target version supports Ruby 3.2.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "bundler"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    # Do not resolve a version published in the last 7 days. A malicious
+    # publish is typically caught and yanked within hours; the delay is
+    # margin against detection lag rather than against the attack itself.
+    cooldown:
+      default-days: 7
+      semver-major-days: 21
--- a/Gemfile
+++ b/Gemfile
@@ -36,5 +36,8 @@ gem "html-proofer", '>=3.3.1'
 gem 'front_matter_parser', '~> 0.1.1'
 gem 'rake'
 
-# Rel issue: https://github.com/ethereumclassic/ECIPs/pull/308#issuecomment-618044919
-gem 'faraday', '~> 0.17.3'
+# Held below 2.0: sawyer (via octokit) requires faraday (> 0.8, < 2.0).
+# 1.10.6 is the lowest release in that range clearing CVE-2026-54297 (recursion
+# DoS) and CVE-2026-25765 (SSRF), which affect the previous 0.17.3 pin.
+# Original 0.17.3 pin: https://github.com/ethereumclassic/ECIPs/pull/308#issuecomment-618044919
+gem 'faraday', '~> 1.10.6'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ GEM
   remote: https://rubygems.org/
   specs:
     Ascii85 (2.0.1)
-    activesupport (8.1.1)
+    activesupport (8.1.3.1)
       base64
       bigdecimal
       concurrent-ruby (~> 1.0, >= 1.3.1)
@@ -15,7 +15,7 @@ GEM
       securerandom (>= 0.3)
       tzinfo (~> 2.0, >= 2.0.5)
       uri (>= 0.13.1)
-    addressable (2.8.8)
+    addressable (2.9.0)
       public_suffix (>= 2.0.2, < 8.0)
     afm (1.0.0)
     async (2.35.0)
@@ -33,7 +33,7 @@ GEM
     coffee-script-source (1.12.2)
     colorator (1.1.0)
     commonmarker (0.23.12)
-    concurrent-ruby (1.3.6)
+    concurrent-ruby (1.3.8)
     connection_pool (3.0.2)
     console (1.34.2)
       fiber-annotation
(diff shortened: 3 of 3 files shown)


IstoraMandiri:
It would be great if we could get build previews on this repo.

These kinds of dep bumps can silently break things visually with SSGs.

IstoraMandiri:
I manually confirmed that this does build. Approving.

IstoraMandiri (approved):

PR #584, opened 2026-08-27
spec(ecip-1111): update to post implementation draft

chris-mercer:
Updates ECIP-1111 to the post-implementation draft, reflecting the specification as built rather than as originally proposed.

Title is now **Base Fee Market and Redirection**, which states what the proposal specifies.

### DIFF
--- a/_specs/ecip-1111.md
+++ b/_specs/ecip-1111.md
@@ -199,7 +199,7 @@ Activation blocks are to be finalized through open coordination among client imp
 - **Mordor Testnet**: `TBD`
 - **ETC Mainnet**: `TBD`
 
-**These blocks are recorded here and in no other document.**
+**This ECIP is the only document in which these blocks are recorded.**
 
 **Cross-client agreement is evidence that this specification is unambiguous only where the implementations do not share a state-transition codebase.** Before this ECIP advances to `Final`, the vectors above MUST be executed on at least two implementations whose state-transition code derives from unrelated upstreams, and both the vectors and their results MUST be published as data a third party can re-execute.
 
--- a/_specs/ecip-1111.md
+++ b/_specs/ecip-1111.md
@@ -1,11 +1,10 @@
 ---
 lang: en
 ecip: 1111
-title: Olympia EVM and Protocol Upgrades
+title: Base Fee Market and Redirection
 status: Draft
 type: Standards Track
 category: Core
-requires: 1112
 author: Cody Burns (@realcodywburns), Chris Mercer (@chris-mercer)
 created: 2025-07-04
 discussions-to: https://github.com/orgs/ethereumclassic/discussions/530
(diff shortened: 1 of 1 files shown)


_specs/ecip-1111.md
```diff
@@ -14,476 +13,223 @@ license: CC0-1.0
...
-
-This modular approach allows ECIP-1111 to remain a single, isolated consensus change, consistent with ECIP-1000. It enables the network to adopt EIP-1559 and EIP-3198 immediately, without introducing:
-
-- Dependencies on governance readiness  
-- Coordination requirements beyond client activation  
-- Any operational or administrative prerequisites  
+**The second line is the whole of the change**, including EIP-1559's own base-fee adjustment arithmetic, which is unchanged.
 
-Activation of governance control over the Treasury is intentionally specified outside of ECIP-1111. It is defined in:
+**The credit is per transaction. Implementations MUST NOT compute it as a single block-level product at finalization**, and MUST NOT do both. `block.gasUsed × block.baseFeePerGas` yields the same total for the block, so the two rules agree on the state root at a block boundary and disagree everywhere inside one: under this ECIP, transaction _N+1_ observes a Vault balance that already carries the base fees of transactions _1..N_ of the same block, and that balance is readable by `BALANCE` and `SELFBALANCE`.
```
realcodywburns: SELFBALANCE (EIP-1884) only returns the balance of the currently executing contract; it can't read an arbitrary external address. Since the vault holds no code (the spec is explicit that nothing executes there), no contract could ever invoke SELFBALANCE against it. Only BALANCE(SOVEREIGNTY_VAULT), called from some other contract, applies. 

... and that balance is readable by  `BALANCE`

_specs/ecip-1111.md
```diff
@@ -14,476 +13,223 @@ license: CC0-1.0
...
+- **Mordor Testnet**: `TBD`
+- **ETC Mainnet**: `TBD`
 
-- **State-Transition Accuracy** — Clients MUST correctly compute and redirect the `BASEFEE` amount to the Treasury. Incorrect calculation or transfer risks:
-  - Rounding inconsistencies  
-  - Ledger imbalance  
-  - Block execution mismatches  
-- **Test Vector Updates** — Test vectors that assume `basefee` burning MUST be adapted to validate Treasury accumulation behavior.
-- **Cross-Client Alignment** — All clients MUST produce identical results for `basefee` calculation, transaction processing, and Treasury accumulation.
+**These blocks are recorded here and in no other document.**
```
realcodywburns: add:
MIN_BASE_FEE Step Changes
A future hard fork that changes MIN_BASE_FEE applies the new value as an immediate step at the activation block, with no EIP-1559-style adjustment ramp. This is consistent with the fork block itself, where INITIAL_BASE_FEE is applied as a discontinuous jump from no base fee to 1 gwei. Both are consensus parameter changes gated by hard fork, and both take effect at a single block by design. Smoothing a floor change would introduce transition-state tracking not required by any other rule in this ECIP, contradicting its minimal-surface intent. This is an accepted design property, not an omission.

realcodywburns (dismissed):
Core mechanism (per-tx crediting, floor clamp, chain-config vault) is sound.

 No blocking issues.

Line 64: SELFBALANCE claim is wrong (no code at the vault, so it can't apply) — corrected to BALANCE only.
MIN_BASE_FEE step changes on future floor updates aren't addressed in Security Considerations. Consistent with the fork block's own discontinuous jump; no smoothing needed, just worth documenting as a choice rather than a gap. Note attached.
Verified ref 5 (Foundry/alloy-rs claim): accurate. Chain 61 is unregistered, is_legacy() returns false, both cited call sites match.

Scope decoupling from 1112–1114 is intentional, not an error; flagging since it shifts who governs the vault later.

_specs/ecip-1111.md
chris-mercer: Corrected in 0134a2a — the sentence now ends at `BALANCE`, which also matches the Test Cases row that already specified BALANCE only.

One refinement on the reasoning, since it matters for ECIP-1112: the address does hold code. This ECIP already allows for that — §"Consensus-Layer Redirection" notes the destination "may hold a contract with a payable fallback" — and #585 specifies a `TransparentUpgradeableProxy` there. What holds is narrower: no EVM code runs as part of the credit, and `SELFBALANCE` returns only the executing contract's own balance, so it cannot be how an external observer reads the Vault mid-block. `BALANCE` is the right opcode for the claim this sentence makes, so the removal stands.

_specs/ecip-1111.md
chris-mercer: Added to Security Considerations in 0134a2a, text as written. One change: the heading reads "Step Changes to the Minimum Base Fee" rather than "MIN_BASE_FEE Step Changes", since no heading in this document carries a code identifier. Constant names are backticked in the body to match the rest of the spec. Happy to switch to your exact wording if you prefer it.

_specs/ecip-1111.md
```diff
@@ -14,476 +13,227 @@ license: CC0-1.0
...
```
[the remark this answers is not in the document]
chris-mercer: Reworded to "This ECIP is the only document in which these blocks are recorded." The sentence is an anti-duplication rule — the activation blocks must not be mirrored into another spec where the two could drift — rather than a claim that they are already fixed. The TBD values are filled in when coordination concludes.

_specs/ecip-1111.md
```diff
@@ -14,476 +13,227 @@ license: CC0-1.0
...
 
-#### Testnet Coordination & Implementation Status
+##### The Destination Address
 
-At the time of writing, prototype implementations of EIP-1559, EIP-3198, and consensus-layer `BASEFEE` redirection are under active development within the Ethereum Classic client ecosystem.
+`SOVEREIGNTY_VAULT` is a chain-configuration parameter holding one address per network, carried in the chain specification each participating client ships — the same configuration surface §Rationale shows Nethermind and Erigon already exposing for this purpose. Changing it therefore costs a coordinated client release, and no on-chain party can change it at all. [ECIP-1112](https://ecips.ethereumclassic.org/ECIPs/ecip-1112) specifies what is deployed at it and how its value is published.
```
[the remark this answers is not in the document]
chris-mercer: Fixed in #585, which rewrites ECIP-1112 around the chain-configuration model. The deterministic-address text is gone; §"The Address" now defers to this ECIP's §"The Destination Address" for the parameter and takes only the publication obligation. #584 should merge first, since #585 cross-references two sections that land here.

IstoraMandiri:
This seems like a big rewrite that is difficult to follow from just a diff. 

Can we get a TLDR on what the new proposal is?

IstoraMandiri (approved):

PR #585, opened 2026-09-01
spec(ecip-1112): update to post implementation draft

chris-mercer:
Updates ECIP-1112 to the post-implementation draft, reflecting the specification as built rather than as originally proposed.

Title is now **Sovereignty Vault**, matching the term ECIP-1111 uses for the chain-configuration parameter holding the address.

The account is specified by three properties readable from chain state rather than by a named contract at a named release, so changing the deployed release is a deployment decision rather than an amendment here.

Cross-references ECIP-1111 §"The Destination Address" and §"Test Cases", which land in #584.

### DIFF
--- a/_specs/ecip-1112.md
+++ b/_specs/ecip-1112.md
@@ -1,7 +1,7 @@
 ---
 lang: en
 ecip: 1112
-title: Olympia Treasury Contract
+title: Sovereignty Vault
 status: Draft
 type: Standards Track
 category: Core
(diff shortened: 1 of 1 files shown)


IstoraMandiri (approved):

chris-mercer:
Noting the simple "WIP" PR title check continues to hang on PRs blocking checks from passing. Blocking merge. Other PRs this check was re-triggered via changing the title. That does not appear to be unsticking this PR.

image

Issue #587, opened 2026-09-11
Preview Builds


IstoraMandiri:
Similar to the EthereumClassic.org website, we should implement preview builds for pull requests to ensure that things look as they should before merging into main.

This can be implemented quite easily with the existing Netlify workflow.

We'll implement this shortly.

Ref #583

PR #588, opened 2026-09-11
Netlify deploy previews

IstoraMandiri:
Adds a Netlify config so every pull request gets a deploy preview, without Netlify ever publishing the production site (which stays on GitHub Pages). The build runs the same Jekyll command as CI, is skipped for all contexts except deploy-preview, installs gems explicitly so CLI-triggered builds work, and pins Ruby 3.2.9 plus Bundler 2.4.19 because the current lockfile's Bundler 2.1.4 crashes on Ruby 3.x. The Bundler pin can be dropped once #583 lands.

Ref #587


diega (approved):
This is a great feature. Thanks!

